PR #6, opened 2026-07-24
feat(skills): canonical repo-root skills, plugin-installable + shipped on npm

joelgwebber:
## What

Make the `sightmap-authoring` / `sightmap-browser` skills a first-class, standalone skill set instead of being buried in the Go module, and ship them through three channels from one source.

## Why

Downstream tools (Subtext) want a Subtext user to get the sightmap skills implicitly, and the skills are useful on their own as a browser-use agent. Today they only reach a user via `sightmap skills install` (embedded in the binary).

## Changes

- **Canonical skills move to repo-root `skills/`** (Layout 2). `go:embed` can't reach outside the `go/` module or follow symlinks, so a committed copy is generated into `go/skills/` by a new `go/skills/internal/gen` tool wired through `//go:generate` in `embed.go`. Same generate-and-commit pattern as the docs schema page.
- **Standalone plugin:** root `.claude-plugin/{plugin.json,marketplace.json}` + `.codex-plugin/plugin.json`, skills-only (no MCP server). Installs like any plugin: `/plugin marketplace add sightmap/sightmap` then `/plugin install sightmap@sightmap-marketplace`.
- **Ships on npm:** `skills/` added to `@sightmap/sightmap`'s `files`; `build-npm-packages.mjs` copies the corpora into the meta package, so downstream consumers vendor them from a pinned version. Verified skills land in the tarball via `npm pack --dry-run`.
- **Skill text:** detect-then-install framing; `npm` baseline, `npx` for one-offs only.
- **Docs:** `README.md`, `go/README.md` (skill links now point at `../skills/`), `AGENTS.md`.
- **CI:** `go` workflow now also triggers on `skills/**` and fails on embed drift (`git status --porcelain`, so untracked files are caught too).

## Validation

`gofmt` clean, `go build ./...`, full `go test ./...` (incl. `browser`) green; `go generate` idempotent and byte-identical to canonical; `sightmap skills install` smoke-tested.

## Follow-ups (not in this PR)

- Cursor/Gemini manifests were deliberately skipped — a skills-only, MCP-less plugin doesn't map cleanly onto Gemini's MCP-centric extension schema; hold pending confirmation.
- Plugin manifest versions are hardcoded (`0.13.1`) with no tag-sync yet.

After merge: cut a tag > 0.13.1 (suggest `v0.14.0`) to publish the skills in `@sightmap/sightmap`. No new trusted-publisher registration (reuses the existing package).


jurassix (approved):
installs worked for me

PR #94, opened 2026-07-29
Fold changesets + release into one workflow, drop the manual tag push

jurassix:
## Summary
- Merges `.github/workflows/changesets.yml` (Version Packages PR) and the old tag-triggered `release.yml` (goreleaser + npm publish) into a single `release.yml` that runs on every push to `main`.
- Once the Version Packages PR is merged (no changesets left to consume), the same workflow now tags the release itself and pushes the tag, then runs goreleaser and publishes the npm packages, closing the loop that previously required a human to run `git tag v<version> && git push origin v<version>`.
- The release step is idempotent: it no-ops if the current `go/npm/package.json` version is already tagged, so ordinary pushes to main (docs-only merges, etc.) are harmless.
- Updates `.changeset/README.md`, `CONTRIBUTING.md`, and `AGENTS.md` to describe the new flow; `maintainers/releasing.md` is untouched since its manual tag step is for the separate spec-versioning process, not this package release.

Mirrors the pattern already proven in `fullstorydev/subtext`, and gives `fullstorydev/subtext-cli` (companion PR) the same shape.

## Test plan
- [ ] `actionlint` passes on the new workflow (done locally)
- [ ] Merge a PR with a changeset, confirm the "Version Packages" PR opens with the expected `go/npm/package.json` bump, `go/npm/CHANGELOG.md`, and synced plugin manifests
- [ ] Merge the Version Packages PR, confirm the workflow tags, runs goreleaser, and publishes all `@sightmap/*` npm packages without manual intervention
- [ ] Confirm `npx @sightmap/sightmap@<version> --version` works post-publish


joelgwebber (approved):
LGTM, and nice to eliminate that manual step. I'll file a follow-up issue for a way we can make it more atomic (to avoid orphaned tags/releases on failure), but that's not a blocker.

PR #97, opened 2026-07-29
SEP-0004: authored classification tags via tags[] (Component, View, Request)

jurassix:
## Summary

Proposes an optional `tags: string[]` field on `Component`, `View`, and `Request` entries —
open-vocabulary classification labels (e.g. `defect`), distinct from identity. Each entity
type already has its own rule for resolving *identity* when multiple definitions could apply
to the same match (nearest-enclosing wins for components, most-specific-route wins for
views, all-matches-apply for requests). Tags deliberately bypass whichever rule applies:
resolution is a **union across every applicable definition**, so a broad, tagged definition
is never shadowed by a narrower, untagged one that wins identity.

Views turned out to have the identical shadowing risk that originally motivated this for
components — same structural shape, just route-specificity instead of DOM ancestry.
Requests need no new resolution logic at all, since "all matches apply" was already the
rule; included for completeness/symmetry.

Backed by a working reference implementation of the component case (Fullstory's internal
Subtext session-review signal pipeline); the view and request cases are new in this SEP,
not yet implemented anywhere.

## What's in this PR

- `spec/seps/0004-component-tags.md` — the full proposal: motivation, shape, JSON Schema
  diff, semantics/conformance (MUST/SHOULD/MAY) per entity type, 5 alternatives considered,
  migration, open questions.
- `spec/v1/sightmap.schema.json` — `tags` added to `$defs.component`, `$defs.view`, and
  `$defs.request`, landed alongside the draft (matching SEP-0002's precedent).
- `spec/v1/schema.md` — `tags` row in all three field tables, plus a new `## Tags` section
  documenting the per-entity resolution rule and a new Conformance bullet.
- `spec/conformance/017-tags.fixture/` + a row in `conformance/README.md`, covering all
  three entity types.

## Verification

Full conformance + example suite reruns clean: 29 fixtures + 5 examples, 0 failures.

## Open questions (see SEP doc)

- Structured/valued tags (`{name, value}`) — deferred, no current demand.
- Console/exception classification — no matchable entity for it today; separate SEP if needed.
- `fmt`-level canonical sorting/dedup for `tags[]`, matching `dependencies[]`.
- Tag namespacing/collision convention for large multi-team corpora.

Field name (`tags` vs. `labels`/`classification`) is settled — keeping `tags`.


joelgwebber (approved):
LGTM

PR #98, opened 2026-07-29
ci(docs): make schema-sync drift self-explanatory + opt-in hook

joelgwebber:
## Why

`docs/reference/schema.md` is generated from the canonical `spec/v1/schema.md` and checked in (Mintlify has no build step), guarded by a CI drift gate. When a PR edits the spec but forgets to regenerate the page, the gate failed with a bare `git diff --exit-code` — a wall of diff and no hint about the remedy. That made an easy-to-fix mistake look like a broken sync process.

## What

- **`docs.yml`** — on drift, print the exact fix (`npm run sync-docs` / `node docs/scripts/sync-spec.mjs` and commit). Mirrors the existing skills-embed drift gate in `go.yml`, which already does this.
- **`package.json`** — add a `sync-docs` script so there's one canonical, memorable regenerate command that CI, the hook, and humans all reference.
- **`.githooks/pre-commit`** — opt-in hook (enable once with `git config core.hooksPath .githooks`) that regenerates and stages the schema page, but only when its inputs (`spec/v1/schema.md` or the generator) are staged, so ordinary commits stay fast. Bypassable with `--no-verify`.
- **`CONTRIBUTING.md`** — document the generated-then-checked-in artifacts and how to enable the hook.

## Notes

The CI drift gate remains the source of truth — it's the only mechanism that can't be bypassed and that works on fork PRs. This PR just makes its failure self-explanatory and gives contributors an easy local guardrail. No auto-commit-from-CI (fork PRs get a read-only token, and bot-pushed commits don't retrigger checks).

Docs/infra-only; no changeset (doesn't touch the published `go/` package).

### DIFF
--- a/.githooks/pre-commit
+++ b/.githooks/pre-commit
@@ -1,0 +1,26 @@
+#!/bin/sh
+# Opt-in pre-commit hook. Enable once with:
+#
+#   git config core.hooksPath .githooks
+#
+# Keeps generated-then-checked-in artifacts from drifting locally so you don't
+# discover it in CI. Currently guards the docs schema page, which is generated
+# from the canonical spec (see docs/scripts/sync-spec.mjs).
+#
+# Only regenerates when the inputs are staged, so ordinary commits stay fast.
+# Bypass with `git commit --no-verify` if you ever need to.
+
+set -e
+
+repo_root=$(git rev-parse --show-toplevel)
+cd "$repo_root"
+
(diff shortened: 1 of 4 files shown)


joelgwebber:
This should make the doc-generation foot-gun a little harder to shoot yourself with. Hopefully we can come back and put it all a bit more on the rails later.

jurassix (approved):

Issue #47, opened 2026-07-26
`requests:` definitions are not consumed


joelgwebber:
A `requests.yaml` validates but has no effect: request definitions are not loaded into the model, and captured network entries are never matched against or named from them.

`requests:` should be wired into the corpus and used to name/annotate network traffic in the devtools output.

chiplay:
Field confirmation from the third-party evaluation against Twenty CRM (chiplay/twenty#1, `SIGHTMAP-EVALUATION.md` defect **D6**, ranked #6 of 14): their corpus shipped 3 `requests:` entries (GraphQL + auth endpoints) that had no observable effect — names, payload docs, and request-level `memory` never surfaced in `network list`/`network get`. Their summary: "a third of the spec is dead weight in the only implementation." Also noted: `:param` request-route normalization passes at validate level (claim C26) but there is no runtime consumption for it to matter, and request-level memory is the one memory placement that never surfaces anywhere (claim C10 fails for requests only). The eval's builder recommendation: implement request-entry surfacing in `network get`, or cut `requests:` from the pitch until it does something.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #113, opened 2026-07-31
SEP-0006: a console/exception `messages:` entity

jurassix:
Declaration and validation layer for [SEP-0006](https://github.com/sightmap/sightmap/blob/main/spec/seps/0006-message-entity.md). Matching is not implemented; `spec/v1/schema.md` says so explicitly.

Stacked on #112.

## Rebased onto the Def convention

`MessageDef` in `go/sightmap/message.go`, replacing `match.Message`, with a JSON wire tag on `Corpus.Messages` so the entity survives serialization into the uploaded corpus. The previous version had no wire tag, so messages would have vanished from the payload.

## `level: ERROR` never matched an exception

SEP-0006's central argument for one entity rather than two was that `level: ERROR` also covers exceptions. It does not. The reference collector labels `Runtime.exceptionThrown` records `level: "exception"`, a distinct value, so under the case-insensitive-exact rule this PR ships, `level: ERROR` could never match one.

The argument survives in stronger form: origin is carried *as a level value*, so no `kind:` discriminator is needed, but a corpus that wants exceptions writes `level: EXCEPTION`. The SEP is amended to say that, and `spec/v1/schema.md` now documents the full emitted vocabulary (`log`, `debug`, `info`, `warn`, `error`, `exception`).

## Validation

| Code | Severity | Condition |
|---|---|---|
| `missing-name` | error | empty name (matching the component and view checks) |
| `message-regex-invalid` | error | `message` does not compile |
| `merge-collision-message` | warning | two entries share a name |
| `message-conflict` | warning | two entries can match the same record, where decidable |
| `message-level-unknown` | lint | a level outside the emitted vocabulary |

`message-regex-invalid` compiles the pattern at validation time. `validateComponent` already does exactly this for author-written selectors three functions away, and the corpus was otherwise storing a regex nobody had proven was a regex.

`merge-collision-message` is load-bearing for #114. `checkSignals` resolves a `ref:` by counting distinct entity *kinds*, so two messages sharing a name collapsed to one kind: neither `signal-ref-unresolved` nor `signal-ref-ambiguous` fired, and the ref resolved silently to an ambiguous pair. Catching the duplicate at its source is what prevents that state.

`message-conflict` covers only the statically decidable overlap (same `level`, identical or absent `message`). Whether two *different* regexes can both match some record is not decidable in general.

`message-level-unknown` is a lint rule rather than a validation error because `level` is deliberately open. The realistic trap is `WARNING`, CDP's own spelling, which the capture normalizes to `warn`, so a corpus copied from CDP output matches nothing and says nothing.

`field-type-invalid` now covers message fields, so `message: 404` and `level: 500` reject in Go as ajv already rejected them. Parity verified on `404`, `500`, `true`, `null`, missing name, and empty name.

## Resolves SEP-0006's open question 3

The SEP explicitly left the ambiguity diagnostic code to its implementation PR. Resolved as `merge-collision-message` plus `message-conflict` for the static half, with the runtime half staying a consumer obligation. That second conformance MUST had been dropped from the earlier draft without being carried into `schema.md`; it is restored.

Verified: `go build`/`vet`/`test ./...`, `npm run validate:conformance`, `npm run validate:examples`, docs sync guardrail.

### DIFF
--- a/.changeset/message-entity.md
+++ b/.changeset/message-entity.md
@@ -8,7 +8,7 @@ Add a top-level `messages:` entity ([SEP-0006](https://github.com/sightmap/sight
 
 New validation:
 
-- `message-regex-invalid` (error) compiles `message` at validation time, matching how component selectors are already checked. The corpus no longer stores a pattern nobody has proven is a pattern.
+- `message-regex-invalid` (error) compiles `message` at validation time, matching how component selectors are already checked. The corpus no longer stores a pattern nobody has proven is a pattern. The dialect is pinned to **RE2** (Go `regexp` / the `re2` npm package for JS): a linear-time syntax with no backreferences or lookaround, so authoring-time validation and runtime matching agree across SDKs.
 - `merge-collision-message` (warning) reports a duplicated name. This one is load-bearing for SEP-0007: `ref:` resolution counts distinct entity kinds, so two messages sharing a name collapse to one kind and the ambiguity check never fires.
 - `message-conflict` (warning) reports two entries that can match the same record, where that overlap is statically decidable: same `level`, identical or absent `message`.
 - `message-level-unknown` (lint) catches a level outside the emitted vocabulary. The realistic trap is `WARNING`, CDP's own spelling, which the capture normalizes to `warn`.
--- a/.changeset/message-entity.md
+++ b/.changeset/message-entity.md
@@ -1,0 +1,19 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Add a top-level `messages:` entity ([SEP-0006](https://github.com/sightmap/sightmap/blob/main/spec/seps/0006-message-entity.md)): named console-output and exception patterns, matched by `level` and a `message` regex. This gives console activity what `requests:` gives network activity, a named entity the rest of the corpus can reference by name.
+
+**An uncaught exception arrives as `level: exception`, not `level: error`.** The reference capture emits `log`, `debug`, `info`, `warn`, `error`, and `exception`, so `level: ERROR` matches `console.error` and does not match an exception. SEP-0006 still needs no `kind:` discriminator, because the origin is carried as a level value, but a corpus that wants exceptions has to name them.
+
+New validation:
+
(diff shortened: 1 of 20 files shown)


joelgwebber:
Pushed a commit covering the fixture-execution gap and the regex dialect:

**Fixture execution.** `019-messages`'s `expected.json` asserted `diagnostics: []`, but `sightmap validate` actually emits a `message-conflict` warning on it — and nothing executed the fixture's `cases` (the JS runner only schema-validates; the Go match runner reads `go/conformance/fixtures`), so the false assertion was invisible. Reworked `019` into a single intentional `message-conflict` and corrected `expected.json`, and added `go/sightmap/spec_conformance_test.go` to run the reference validator over the `*-messages` fixtures and assert their `validate`-case diagnostics exactly. Scoped to message fixtures deliberately (a whole-suite executor also needs `match`/`lint`/`explain`).

**RE2.** Pinned the `message` regex to RE2 (Go `regexp` / the `re2` npm package for JS) in `schema.md`, the JSON Schema, and SEP-0006 — a linear-time dialect (no backreferences/lookaround) so authoring-time validation and runtime matching agree across SDKs, and it's what the Go reference SDK compiles anyway.

**Two follow-ups, not this PR:**
1. Runtime matching is unimplemented (this PR is declaration + validation only). Matching console/exception records against `messages:` — CDP derivation of `Runtime.exceptionThrown`/`consoleAPICalled`/`Log.entryAdded`, a record type, the predicate, plus the deferred "surface ambiguity on live multi-match" MUST — is the direct console sibling of #47. Filed as #170.
2. `017-tags` has the same class of bug this PR fixes: view-level `tags:` trips `unknown field "tags"` because `viewFields` omits `tags`, yet SEP-0004, `schema.md`, and 017's own `expected.json` all treat view tags as valid. Real SEP-0004 gap, and the reason the whole-suite executor can't be turned on yet.

joelgwebber (approved):

joelgwebber (approved):

PR #115, opened 2026-07-31
test(clitest): declarative CLI cases for agent-facing UX

joelgwebber:
## What

A small [`testscript`](https://pkg.go.dev/github.com/rogpeppe/go-internal/testscript)-inspired harness under `go/clitest/` for exercising the CLI's **behaviour at the points where authoring goes wrong** — malformed corpora, hand-written session files, missing commands, and other silent, generic, or misleading output. Ordinary unit tests check that the code does what the code intends; these check that when the tool is used wrong, it says something that gets the author *unstuck*.

These are plain integration tests: `go test ./clitest/` runs them, and so does CI (they live under `go/`, inside the existing `go test ./...` job). No second harness binary, no new workflow.

## How a case looks

One directory per case under `testdata/cases/<slug>/` with a `case.yaml` (command + expectations) and an optional `fixture/` that becomes the run's cwd. Runs are isolated (temp cwd, isolated `$HOME`). Two features beyond a plain table test:

- **`known_bug` xfail ratchet.** A case marked `known_bug: true` asserts *desired, not-yet-shipped* behaviour, so it passes while the tool still misbehaves and turns **red once the behaviour ships** (prompting you to drop the flag). A regression turns a passing case red; a fix turns a known-bug case red until it's claimed. The dataset stays honest in both directions and doubles as an executable backlog.
- **`issue:` as the join key.** The one durable link out of a case is its public GitHub issue number — the shared coordinate across tracking layers, so there's no parallel failure-mode/PR taxonomy. Decontextualized provenance lives in the issue, not the case.

See `go/clitest/README.md` for the full format and the intake ritual for adding cases from a new session.

## Contents

16 cases seeded from real authoring sessions: 13 `known_bug` (silent view-file schema traps, unhelpful `report`/`--help`/session-file diagnostics, missing `init`/`register`, tooling-file warnings, capped `sel-probe` count) and 3 `known_bug: false` guards that pin the happy path so new diagnostics don't leak onto correct corpora.

Local run (macOS): the linux-only and manual cases skip; everything else is green.

## Follow-up

Linking each `known_bug` case to its `issue:` and reconciling against the existing tracker (some already have issues; a few need filing) is a deliberate next pass, not part of this structural change.


go/clitest/testdata/cases/report-error-teaches-schema/case.yaml
```diff
@@ -0,0 +1,16 @@
+title: "report's 'no views with URLs' error names the wrong field and never shows the views: shape"
+known_bug: true
```
jurassix, line 3: very nice that we can red/green these as well.

jurassix (approved):
lgtm - could be cool to get this wired up to run with our prs

joelgwebber:
> lgtm - could be cool to get this wired up to run with our prs

Good news -- they actually get run automatically with CI, because they're just Go tests!

PR #154, opened 2026-08-01
ci(changelog): generate the docs changelog entry during the release

chiplay:
## What this changes

Adds `scripts/changelog-entry.mjs` and chains it into `npm run version-packages`, so the docs-site changelog entry for a release is generated from `go/npm/CHANGELOG.md` at release time instead of being written by hand afterward. Adds a `--check` mode, wired into the `docs` workflow, asserting that every released version has an entry.

## Why

The entries in `docs/changelog/entries/` were authored manually, so a release could ship with no entry at all. That's what happened with **0.17.0**: npm was a minor version ahead of docs.sightmap.org until it was noticed (#153 adds the missing entry).

The existing CI guard couldn't catch it. `build-changelog.mjs` + `git diff --exit-code` only proves `changelog.mdx` matches the entry files — with no entry, there's nothing to be stale against, so it passes. And `docs.yml` was path-filtered to `docs/**`, so it never ran on the "Version Packages" PR, which touches only `go/` and the plugin manifests.

The fix follows a pattern already in the repo. `version-packages` is `changeset version && node scripts/sync-manifest-versions.mjs` — the release PR already regenerates derived files (the plugin manifest versions), and `changesets/action` sweeps whatever that command touched into the PR branch. The changelog entry was simply the one derived artifact never wired into that hook:

```
changeset version && sync-manifest-versions.mjs && changelog-entry.mjs && build-changelog.mjs
```

The entry now lands in the same "Version Packages" PR as the bump that caused it — reviewable before merge, atomic with the release, and drift becomes structurally impossible rather than something to detect afterward.

### Prose is copied, not rewritten

Changesets output is already the published wording. Only the scaffolding is stripped — the `### Patch Changes` heading and the `- <hash>:` prefixes — reproducing how the entries were formatted by hand (one paragraph per changeset, nested bullets dedented to column 0).

Verified two ways:

- Regenerating **0.17.0** from `main` produces a file **byte-identical** to the hand-written entry in #153, tag included.
- Generating **0.17.1** from the in-flight release PR (#147) — three changesets, nested bullets, a trailing paragraph — reproduces the multi-changeset convention established by `0.15.7`.

A full rehearsal of the post-`changeset version` half of the command against #147's tree produces the correct manifests, entry, and rebuilt page, and is idempotent: re-running leaves an existing entry untouched, so a hand-edited entry survives.

### Two fields stay a reviewer's call

The script prints both on stdout rather than pretending to be sure:

- **`tags`** — `["go"]` vs `["cli"]` is a judgment about audience that the prose doesn't state. Guessed from library-surface phrasing; it got 0.17.0 and 0.17.1 right, but it is a heuristic.
- **`date`** — the day the Version Packages PR is generated. Skews by a day if that PR sits unmerged across a date boundary.

Both are trivially fixable in the release PR, which is the point of generating them there.

## Area

- [ ] Spec (`spec/`)
- [ ] Go implementation / CLI (`go/`)
- [x] Docs site (`docs/`)
- [ ] Marketing site (`web/`)
- [x] Maintainer / infrastructure

## Type of change

- [ ] Bug fix
- [ ] Docs / wording / typo
- [x] Feature
- [ ] Spec clarification (no semantic change)
- [ ] Spec change (requires an accepted SEP — link below)
- [ ] Example or conformance fixture added/updated

## Linked issues or SEPs

Follows #153, which adds the 0.17.0 entry this would have generated automatically.

## Merge order

**Merge #153 first.** Until it lands, `main` has no 0.17.0 entry, so this PR's new coverage check fails — the check correctly reporting the real gap it was written to catch. It goes green once #153 merges; no changes needed here.

Worth noting #147 ("Version Packages", 0.17.1) is open now. If it merges before this, 0.17.1 also ships without an entry and will need one added by hand — the last release that would.

## Checklist

- [x] I read [`CONTRIBUTING.md`](../CONTRIBUTING.md)
- [ ] ~~Every commit on this branch is signed off~~ — **not done, needs a maintainer.** The commit is authored by `Claude <noreply@anthropic.com>`; a DCO sign-off is a certification by a human contributor, so it isn't mine to make. Please `git commit --amend -s` (or squash-merge with sign-off) before merging. The same applies to #153.
- [x] I kept this PR focused on a single concern
- [ ] If this changes the spec, there is an accepted SEP linked above — n/a
- [ ] If this changes the JSON Schema, I updated the schema file, `spec/v1/schema.md`, and any affected examples — n/a
- [x] Relevant checks pass locally — generator and both `--check` modes verified; no Go or site build affected

---
_Generated by [Claude Code](https://claude.ai/code/session_0138rvrzy5eip9a6x8C2UZAs)_

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,11 +1,14 @@
 name: docs
 
+# go/npm/CHANGELOG.md is in the path filter so the changelog-coverage check
+# below runs on the "Version Packages" PR, which touches no docs/ files but is
+# exactly where a release without an entry would slip through.
 on:
   push:
     branches: [main]
-    paths: ['docs/**', 'spec/v1/**', '.github/workflows/docs.yml']
+    paths: ['docs/**', 'spec/v1/**', 'go/npm/CHANGELOG.md', 'scripts/changelog-entry.mjs', '.github/workflows/docs.yml']
   pull_request:
-    paths: ['docs/**', 'spec/v1/**', '.github/workflows/docs.yml']
+    paths: ['docs/**', 'spec/v1/**', 'go/npm/CHANGELOG.md', 'scripts/changelog-entry.mjs', '.github/workflows/docs.yml']
 
 permissions:
   contents: read
@@ -42,6 +45,13 @@ jobs:
         run: |
           node docs/scripts/build-changelog.mjs
           git diff --exit-code docs/changelog.mdx
+
+      # The drift check above only proves the page matches the entries — it
+      # passes just fine when a release has no entry at all, which is how the
+      # docs site fell a version behind npm. This asserts coverage instead:
+      # every released version in go/npm/CHANGELOG.md has an entry.
+      - name: Every release has a changelog entry
+        run: node scripts/changelog-entry.mjs --check
 
       - name: Install Mintlify CLI
         run: npm i -g mint
--- a/docs/changelog.mdx
+++ b/docs/changelog.mdx
@@ -8,6 +8,60 @@ rss: true
     Add an entry in changelog/entries/, then run: node scripts/build-changelog.mjs
     Docs: changelog/README.md */}
 
+<Update label="sightmap 0.18.0" description="Released August 5, 2026" tags={["go","cli"]}>
+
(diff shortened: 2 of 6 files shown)


chiplay:
@joelgwebber @jurassix thoughts on this one? Trying to keep our changelog in sync on the docs site, but not sure this is the best approach. https://docs.sightmap.org/changelog

jurassix (approved):
lgtm

PR #161, opened 2026-08-03
match: rename SightmapComponent/SightmapMatch -> ComponentDef/ComponentMatch

joelgwebber:
## What

Rename the rule-side corpus types to the `Def` convention:

- `match.SightmapComponent` → `match.ComponentDef`
- `match.SightmapMatch` → `match.ComponentMatch`

Pure mechanical rename across the module (~39 files, 217/217); **no behavior change**. Done as a word-boundary rewrite, so the domain `buildSightmapMatchMap*` helpers (which build node→name maps, not the type) are intentionally left alone.

## Why

Establishes the vocabulary for the corpus/runtime data model: a `ComponentDef` *matches* a `Component`; a `RequestDef` classifies a `Request`. Per-entity result names (`ComponentMatch`) avoid a generic `match.Match` stutter, and we deliberately don't pre-create `ViewMatch`/`RequestMatch` (view/request matching return their defs directly). Observed log/exception types stay out of `match` — that Family A/B boundary keeps the package bounded.

## Breaking

`match.SightmapComponent` / `SightmapMatch` are exported and consumed downstream (FullStory/lidar). Consumers pick this up on their next dep bump.

## Notes

- NOTE: changeset intentionally omitted pending a semver-bump call (breaking rename).
- First of a two-PR stack; the corpus-wire PR builds on this branch.

### DIFF
--- a/.changeset/def-rename.md
+++ b/.changeset/def-rename.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+**Breaking:** rename the rule-side corpus types to the `Def` convention — `match.SightmapComponent` → `match.ComponentDef` and `match.SightmapMatch` → `match.ComponentMatch`. A `ComponentDef` describes what to match; a `ComponentMatch` is the result of matching one. No behaviour change; callers update to the new names.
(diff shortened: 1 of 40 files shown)


jurassix (approved):
lgtm

PR #162, opened 2026-08-03
sightmap: model requests + RequestsForURL, JSON wire tags, stable pre-order

### LINKED ISSUE #47: `requests:` definitions are not consumed

joelgwebber:
## What

Make the corpus a deterministic, serializable, requests-aware wire form:

- **Model `requests:`** — global + view-scoped `RequestDef`/`Payload`/`Field` on `Corpus`/`View`, parsed by the loader (with `missing-name`/`missing-route` diagnostics for parity with components).
- **`Corpus.RequestsForURL(url, method)`** — route glob + optional method, "all matches apply" per spec; reuses the existing route matcher (`:param`, trailing-slash, `**` all handled).
- **JSON wire tags** on `Corpus`/`View`/`RequestDef` — serializes to the lean wire (`memory`/`globals`/`views`/`requests`); authoring-only `View` fields (`url`/`access`/`snapshots`/`sourceFile`/`stability`) are `json:"-"`.
- **Stable pre-order** — the flatten is already deterministic (lexical file order + DFS declaration order); locked with a regression test so the wire form stays reproducible.

Closes #47 (`requests:` were known to the schema + unknown-field checker but never modeled or consumed).

## Why

Unblocks downstream network-signal enrichment (semantic names on network requests) and gives the corpus a reproducible wire/storage/serve form.

## Notes

- **Stacked on #161** (the `Def` rename) — merge that first; GitHub retargets this base to `main` automatically.
- NOTE: changeset intentionally omitted pending a semver-bump call (additive: new `RequestsForURL` API + wire tags).

### DIFF
--- a/go/sightmap/corpus.go
+++ b/go/sightmap/corpus.go
@@ -4,15 +4,14 @@ import (
 	"net/url"
 	"regexp"
 	"strings"
-	"sync"
 
 	"github.com/sightmap/sightmap/go/match"
 )
 
-// Corpus is the parsed, $ref-expanded, hierarchy-flattened sightmap data, plus a
-// lazily-populated cache of compiled match queries. It is the single operational
-// handle callers hold: load one with Load (or a Loader), then MatchTree a live
-// component tree against it. Safe for concurrent use.
+// Corpus is the parsed, $ref-expanded, hierarchy-flattened sightmap data. It is
+// pure, serializable data: load one with Load (or a Loader) and read it with the
+// ComponentsForURL / ViewForURL / RequestsForURL queries. To match a live
+// component tree against it, build a Matcher with NewMatcher(corpus).
 type Corpus struct {
 	// Memory holds file-level memory entries from every corpus file, in
 	// file-path order. The spec activates these per source file ("applies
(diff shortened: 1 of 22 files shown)


go/sightmap/corpus.go
```diff
@@ -20,15 +20,21 @@ type Corpus struct {
...
 	// Views contains per-route component lists, with $refs expanded and
 	// children flattened.
-	Views []View
+	Views []View `json:"views,omitempty"`
+
+	// Requests is the flat list of globally-defined API request definitions
+	// (from a file-root `requests:` list). View-scoped requests live on each
+	// View instead. Match against an observed network request via
+	// RequestsForURL.
+	Requests []RequestDef `json:"requests,omitempty"`
```
jurassix, line 36: SEP-5,6,7 - reminder that when this lands we should ensure the implementation hits the surface area in this pr.

jurassix (approved):
lgtm

PR #163, opened 2026-08-03
sightmap: extract Matcher engine (Corpus becomes pure data)

joelgwebber:
## What

Extract the matching engine off `Corpus` into a new `Matcher`:

- `Corpus` is now **pure, serializable data** (no cache, no mutex).
- `NewMatcher(corpus)` holds the per-URL compiled-query cache; `MatchTree` / `Components` move onto it.
- Callers build a `Matcher` where they match (one per matching scope, so the cache is shared within a scope). No behavior change.

## Why

Completes the data-vs-engine split: `Corpus` is the wire/storage/serve form, `Matcher` is the engine. Precondition for later serving the `Corpus` wire directly.

## Scope

- **Left alone on purpose:** `cmd/sightmap` serve's `compiledSightmapJSON`. Replacing it changes the wire the browser extension consumes, so that's a separate coordinated client+server change — this PR is engine-only.

## Notes

- **Stacked on #162** (corpus wire format) → #161 (rename). Merge in order; bases retarget automatically.
- NOTE: changeset pending a semver call (breaking: `Corpus.MatchTree`/`Components` now live on `Matcher`).

### DIFF
--- a/.changeset/matcher-engine.md
+++ b/.changeset/matcher-engine.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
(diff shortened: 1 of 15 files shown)


jurassix (approved):
lgtm - pushed a small commit to clean up some dup

PR #167, opened 2026-08-05
sightmap: add stats verb — corpus totals + per-view component/request table

chiplay:
Adds `sightmap stats [--sightmap-dir DIR] [--json]`, so you can see the size of a corpus without reading it.

Counting lives in the library as `Corpus.Stats()` (`go/sightmap/stats.go`), not in the command. It encodes corpus-model knowledge that belongs beside the model it traverses: memory entries live in four places, requests are global plus view-scoped, and components dedupe by first-seen name after `$ref` expansion. A YAML walk that reimplements any of that drifts the moment the model grows. `cmd_stats.go` is a thin shell over it, and the atlas publisher CI is the second intended caller.

Human mode prints totals plus a per-view table. `--json` emits the field set the sightmap/atlas gallery CI consumes, pinned by a unit test:

```json
{"views":2,"components":10,"requests":2,"properties":0,"memory":5,
 "per_view":[{"name":"Checkout","route":"/checkout","components":4,"requests":1}]}
```

Totals can exceed the per-view sum when a corpus has global-only components. `components` counts distinct names; `properties` and `memory` count distinct definitions, so two views that legitimately define different components under the same name both contribute.

`stats` refuses to count a corpus that `sightmap validate` rejects. The loader silently drops definitions it cannot resolve, so an unresolved `$ref` would otherwise report confident, quietly wrong numbers and CI would bless a broken entry. Under `--json` the refusal is still one JSON object on stdout, an `{error, diagnostics[]}` envelope, so a consumer that unconditionally parses stdout always has something to parse.

Verified against a real corpus: the first atlas entry reports 4 views, 28 components, 2 requests, 16 properties, 20 memory notes.

Tests cover the dedupe fold, the published field set, the refusal path, empty and globals-only corpora, and trailing positional arguments. Two of the original clitest cases were pure correctness guards and moved to unit tests per `go/clitest/README.md`, replaced by a case for the refusal, which is the diagnosability moment an author actually hits. `gofmt`, `go build ./...`, `go vet ./...`, and `go test ./...` are green. Docs land in the same PR: `docs/cli/corpus.mdx`, `go/docs/reference/tools.md`, and the authoring skill tables, with every output block pasted from the real binary.

Changeset: `.changeset/stats-verb.md` (minor).

First of two CLI verbs for the community atlas (github.com/sightmap/atlas, `plan/phases/phase-1-cli.md`). The sibling PR #168 touches the same `main.go` dispatch table, so whichever merges second needs a trivial rebase.


chiplay:
## Review findings addressed

All four verified findings fixed, plus the docs, test-coverage, and architecture items from the same review. An adversarial verifier re-ran every reproduction against a rebuilt binary and returned SHIP.

**Architecture.** Counting moved into the library as `Corpus.Stats()` (`go/sightmap/stats.go`). It encodes corpus-model knowledge — that memory lives in four places, that requests are global plus view-scoped, the dedupe-vs-expansion split — so it belongs beside the model it traverses, per ARCHITECTURE.md's tier contract. Not a `RunStats(opts)` twin: no options struct, no CLI flavor, and the package imports only `strings` and `match`. `cmd_stats.go` is now a thin shell. The atlas publisher CI is a second intended consumer.

**Fixes** (each with a reproduction in the verifier's run):
- **Load diagnostics were ignored** — a corpus with an unresolved `$ref` typo reported `components=2`, exit 0, because the loader silently drops the component. `stats` now refuses to count what `validate` rejects. Under `--json` the refusal is still exactly one JSON object on stdout — a `{error, diagnostics[]}` envelope — so a consumer that unconditionally parses stdout always has something to parse.
- **`--json` broke on empty/memory-only corpora** — the empty gate ran before the JSON branch, so a legal memory-only corpus got a human error and exit 1. JSON is now answered first, and memory counts toward non-emptiness.
- **Properties and memory undercounted** — two views defining different local components sharing a name dropped the second's counts (3 properties reported as 1). Totals now key on definition identity; `components` keeps its documented distinct-names meaning.
- **Trailing positional args ignored** — `sightmap stats json` printed the table and exited 0; it now errors.

**Tests and docs.** `runStatsOut(args, io.Writer)` makes the emitted JSON assertable in-process, and a contract test pins the exact published field set. The two clitest cases that were pure correctness guards moved to unit tests per `clitest/README.md`, replaced by a diagnosability case for the new refusal. `stats` is documented in `docs/cli/corpus.mdx`, `go/docs/reference/tools.md`, and the authoring skill tables, with all output blocks pasted from a real binary run. The shared per-view table helper is now used by `report` too — output verified byte-identical across every width path.

Known follow-up, non-blocking: the `overview.mdx` exit-code row for an empty corpus doesn't mention that it only applies in table mode.

jurassix:
This one seems fine to me. The `--json` flag could probably be a global thing for any cmd returning output but lgtm. I think @joelgwebber should take a look before we merge.

jurassix (approved):

PR #168, opened 2026-08-05
sightmap: atlas find/list/add — search the atlas by domain, install without the index

chiplay:
Adds three verbs for the community atlas (github.com/sightmap/atlas):

```
sightmap atlas list                 # browse the catalog
sightmap atlas find <query>         # search by domain, name, category, description
sightmap atlas add <slug>           # install a corpus into .sightmap/
```

## Why search exists

The consumer that matters most is an agent about to automate a site. It has a URL, `squareup.com`, and no way to learn the product is called Square POS. A surface that only accepts a slug someone read off a gallery card is unusable to it. `find` matches slug, name, `domains[]`, `categories[]`, and description, ranking an exact domain match first, and prints each hit's own install command so search hands off to install with nothing to construct.

An empty result is a successful search, so `find` exits 0 with a suggestion to try the product name or a category. `add` with an unknown slug is a failed action and exits 1. Agents branch on that difference.

## Install stopped reading the index

`add` fetches one archive from a URL template and never touches `index.json`. `find` and `list` read the index, cached at `~/.sightmap/atlas/index.json` with a 24h TTL (`--refresh` busts it), following the `~/.sightmap/browsers` precedent.

Separating them means installs keep working through an index outage or a schema change, the index can grow fields without a CLI release, and an agent narrowing a query across three calls hits the network once. It also deletes the reason `files[]`, per-entry `commit` pinning, and raw-URL layout derivation existed. Commit pinning is gone entirely: reproducibility for a corpus is that the YAML is checked into your repo.

## What this replaces

The previous revision installed by fetching each file listed in the index, and made `--force` replace rather than merge. Replace-not-merge was never a requirement. It created the need to delete the target, which produced a bug that deleted a user's `.git`, `.env`, and `src/` and exited 0, which then needed a 158-line allowlist to protect the feature from itself. Refusing a non-empty target and telling the user to delete it is the whole requirement, and atomicity is now just the rename.

Test code drops from 1,940 lines to 1,399: 1,754 lines of install and replace machinery became 364 lines of install tests. Production code grows by 243, because search and the index cache are new capability.

## Security posture, kept

Four things survive the simplification, because archives and a search verb change the threat model rather than removing it.

**Transport policy.** HTTPS-only with a loopback exception, checked before the request is built and re-checked on every redirect hop, capped at 5. One archive fetch still follows redirects, so a 302 to plain HTTP would otherwise install attacker bytes. Covered by a downgrade at hop 3 of 3 and a loopback-redirect control proving redirects are still followed.

**Output escaping.** `find` and `list` print index-supplied names, descriptions, and categories to a terminal, which is more untrusted text than `add` ever printed. Every such string goes through `SafeText` or `safeList`. Search drops an entry only when its *slug* is unusable, since that would print an uninstallable command; a stray byte in a description gets escaped rather than making the site unfindable.

**Caps sized for archives.** 8 MiB on the wire, 32 MiB decompressed via a reader that fails mid-decompression, 4 MiB per file checked against both the header and the bytes that arrive, and 512 members. A compression bomb is small on the wire, so the decompressed cap is the one that catches it, and a test asserts the wire cap would not have. Members that are absolute, contain `..`, escape the target after joining, or are symlinks, hardlinks, or devices refuse the install.

**`schema_version` gating moved rather than disappeared.** `ParseIndex` gates on a standalone decode before entries are unmarshalled, so a restructured v2 index produces "upgrade sightmap" instead of a Go type error. `LoadIndex` reports it from the network read, so a cached failure doesn't pin the user for a day.

## Architecture

`go/atlas` stays a subsystem with a domain-shaped surface (`Index`, `Entry`, `Search`, `Client`, `Install`), not a `RunAtlasAdd(args)` twin, per `go/ARCHITECTURE.md`. The cmd files are thin shells. The command→subsystem table gains the verb group.

## Verification

`gofmt`, `go build ./...`, `go vet ./...`, `go test -count=1 ./...`, and `go test -race ./atlas/... ./cmd/... ./clitest/...` all green. All three verbs demoed end to end against a local atlas, with every documented output block pasted from the real binary. Three clitest cases cover the refusals an agent hits first: missing slug, non-HTTPS index, existing corpus.

Docs: `docs/cli/integrate.mdx`, `docs/cli/overview.mdx`, `docs/reference/go-library.mdx`, `go/docs/reference/tools.md`, `go/atlas/doc.go`, and the authoring skill, which now tells an agent to check the atlas before mapping a site by hand.

## Two things to know before merging

`add` needs a published archive per entry. The atlas repo's `publish.yml` (sightmap/atlas#2) has to emit `entries/<slug>.tar.gz` alongside `index.json`; that URL template is a contract even though it isn't the index schema. Filed against that repo.

The "browse the gallery" pointer in error messages goes to the GitHub repo rather than sightmap.org/atlas, since the gallery is still in review. One constant, `atlas.AtlasURL`, to change when it ships.

Sibling PR #167 touches the same `main.go` dispatch table, so whichever merges second needs a trivial rebase.


chiplay:
## Review findings addressed

All twelve verified findings fixed, plus docs, clitest cases, and the architecture item. An adversarial verifier re-ran every security reproduction against a rebuilt binary — and caught a HIGH-severity regression introduced by one of the fixes, which is also fixed here.

**Architecture.** The wire contract moved into a new `go/atlas` subsystem: index schema, ref/commit resolution, fail-closed validators, transport policy, and the install operation. `cmd_add.go` is now 88 lines of flags-in, format-out. The exported surface is domain-shaped (`Index`/`Entry`/`Source`/`Client`/`Install`), not a `RunAdd(args)` twin — ARCHITECTURE.md forbids commands-as-library. This is what lets sightmap/atlas publish-time CI import the same validators instead of reimplementing them and drifting.

**Security.**
- **Redirects bypassed the transport policy** — a mirror answering `302 → http://evil/x` got its plaintext bytes installed. The policy now re-runs on every hop; a control case proves redirects are still followed.
- **Terminal escape injection** — index-supplied names and suggestion slugs reached the terminal raw. Control bytes are now rejected in validation and escaped at the output boundary; verified no raw `0x1b` on stdout or stderr.
- **No size or count caps** — four caps now enforced (index, per-file, per-entry aggregate, file count), each reproduced with nothing installed.

**Correctness.** `--target ""` no longer writes into the cwd; unpinned fetches follow the ref from `--index` instead of hardcoding `main`; a missing commit warns that files came from a floating ref; GitHub's own `refs/heads/…` raw URL shape now resolves pinned entries correctly; the `schema_version` gate decodes before the shape, so a restructured v2 index gets the upgrade message rather than a JSON parse error; installs are atomic via stage-then-rename; local preconditions are checked before any network I/O; and `--force` replaces rather than merges, so stale files no longer survive.

**The regression that fix caught.** Making `--force` replace-not-merge meant renaming the target aside and deleting it — with nothing constraining what the target could be. `sightmap add shop --target . --force` in a project directory permanently destroyed `.git`, `.env`, `go.mod`, and `src/`, and exited 0. Now bounded two ways: refused by **location** (the cwd, any ancestor of it, home, any ancestor of home, a filesystem root — symlink-resolved on both sides) and by **contents** (an allowlist: every top-level entry must be a corpus YAML file or `views/`, `snapshots/`, `review/`). Allowlist rather than a blocklist of scary names, because a blocklist only refuses the destruction someone anticipated. The backup is also no longer deleted until the post-install load check passes, so a broken atlas entry rolls back and says so. `--force` on a real corpus directory is unchanged.

Also fixed: exported `Client.Fetch` promised HTTPS-only but never checked its first hop — unreachable via the CLI, but the atlas publisher CI is invited to rely on that surface.

**Quality.** One documented mirror-layout rule that works for GitHub raw, Enterprise raw, and a plain copy. Post-install load check so a broken atlas entry is reported as an atlas problem, not the user's. Bounded-concurrency fetches. Slug validated before the index fetch, so the error stops blaming the atlas for a string the user typed. Suggestions simplified to containment.

**Tests and docs.** `go/atlas` at 89% coverage, race-clean; all fixes mutation-checked (reverting each one fails its test). Three clitest cases for the refusals an agent hits first, each asserting no network request was made. `add` documented in `docs/cli/integrate.mdx`, `go/docs/reference/tools.md`, `go/atlas/doc.go`, and the skill tables — including what `--force` refuses to do.

jurassix (commented):
will need to marinate on this one for a bit

jurassix (commented):
This tracks the plan closely: `find`/`list`/`add` under `atlas`, ranking with `RankDomainExact` first, `normalizeDomain` so a pasted URL resolves, exit 0 on no match, `--limit`/`--json`/`--category`/`--refresh`, 24h cache at `~/.sightmap/atlas`, install reads no index, one tarball, stage-extract-load-rename, `--force` gone, and a SKILL.md section with the search-order strategy. The JSON hit even carries `install`, and the skill warns that an entry's `memory:` text lands in your context.

The gaps are all about where the bytes come from.

## 1. The archive doesn't exist, and the URL is the wrong shape

```go
const DefaultArchiveURL = "https://raw.githubusercontent.com/sightmap/atlas/main/entries/{slug}.tar.gz"
```

SPEC.md defines `entries/<slug>/` as a **directory** (README.md, `.sightmap/`, screenshots/). So this expects a committed binary tarball sitting next to a directory of the same name, and nothing in atlas CI produces one. Today `atlas add` cannot install anything.

It's also the worse of the two options. A generated `.tar.gz` per entry, committed to git, churns a binary blob on every content edit. The plan had the site build emit it from the tree #172 already vendors, which is a build artifact rather than a source file.

## 2. Everything points at GitHub instead of sightmap.org

```go
DefaultIndexURL   = "https://raw.githubusercontent.com/sightmap/atlas/main/index.json"
AtlasURL          = "https://github.com/sightmap/atlas"
```

Three consequences:

**Takedown hole.** This is the argument from the #172 review. `removed.yaml` and POLICY.md bind the gallery: a removed entry disappears from `dist/` on rebuild. Pointed at raw.githubusercontent, it stays both findable and installable. One moderation surface only works if both the index and the archive come from our domain.

**Two sources for one document.** #172 already serves `/atlas/index.json` byte-for-byte as vendored. The CLI reading a different copy means they can disagree.

**Wrong destination on a miss.** `AtlasURL` is what `find` prints when nothing matches and what `add` prints on a 404. It sends someone to a repo file tree instead of the gallery.

All three are constant changes:

```go
DefaultIndexURL   = "https://sightmap.org/atlas/index.json"
DefaultArchiveURL = "https://sightmap.org/atlas/{slug}.tar.gz"
AtlasURL          = "https://sightmap.org/atlas"
```

Which makes the real dependency the #172 follow-up: vendor `<slug>/.sightmap/` and have `build-atlas.ts` write the tarball.

## Two nits

**Bidirectional containment fires wide on short fields.** `matchesText` does `Contains(field, want) || Contains(want, field)`. The reverse direction exists so `square-pos-terminal` finds `square-pos`, which is right for slugs. Applied to categories it means querying `position tracking` matches every entry categorized `pos`. Worth limiting the reverse direction to slug and domain, or gating it on a minimum field length.

**`Detail()` drops requests.** `atlas.Stats` carries only `Views` and `Components`, so `find` prints "12 views, 87 components" while the gallery card prints "12 views · 87 components · 23 requests" from the same `index.json`. Same data, two summaries.

## Not a gap

`--source` on `add` takes a URL template, so `sightmap atlas add toast-pos --source https://internal.corp/{slug}.tar.gz` covers the private-corpus case without a separate verb. That closes the thing I was holding open.

jurassix:
@chiplay trying something out on these prs where I'm letting claude review process for me after we go back and forth internally.

chiplay:
Thanks Clint — all four addressed in aa7ad5c, plus one extra flagged below.

## The archive URL

You're right on both counts, and the second one is the part I got wrong. `entries/<slug>/` is a directory in SPEC.md, so `entries/{slug}.tar.gz` expected a committed binary sitting next to a directory of the same name. Nothing produced it, so `atlas add` could not install anything. Your version, where the site build emits it from the tree #172 already vendors, is also the better artifact: no binary blob churning on every content edit.

The takedown argument is the one I should have reached myself. Moderation that binds the gallery but leaves the entry installable from raw.githubusercontent isn't moderation. All three constants now come from the domain we control:

```go
const DefaultIndexURL   = "https://sightmap.org/atlas/index.json"
const AtlasURL          = "https://sightmap.org/atlas"
const DefaultArchiveURL = "https://sightmap.org/atlas/{slug}.tar.gz"
```

All three reasons are in the code comments, the docs, and the changeset, so the next person to consider a convenient GitHub URL finds the argument. A guard test asserts every default resolves to sightmap.org. `--index` and `--source` still override, which is how mirrors and private catalogs work.

This makes #172's follow-up blocking, so I picked it up there: the vendored tree gains `<slug>/.sightmap/`, and `build-atlas.ts` writes the tarball. Until that merges these defaults point at URLs that don't resolve yet. sightmap/atlas#6 is rewritten to match — `publish.yml` vendors the corpus rather than producing archives.

## Bidirectional matching

Narrowed to slug and domain. Name went to forward-only containment as well, which is a judgment call worth stating: the reverse direction asks "is this field a substring of the query," which only makes sense when the field is a canonical identifier someone pastes a longer real-world form of. A name is prose you quote *from*, not extend, and a short one carries the same blast radius you found in categories — an entry named `POS` would answer every query that happens to spell it. `square pos terminal` still lands through the slug and description ranks.

Tests lock both directions: `square-pos-terminal` still finds `square-pos`, and `position tracking` now matches neither a `pos` category nor a `POS` name, through the query and through `--category`.

## Requests in Detail()

Added to `Stats`, so `find` and the gallery card now read the same: `12 views, 87 components, 23 requests`. An index predating the field drops the clause rather than printing `0 requests`, since that would assert a corpus maps none. Test parses a requests-free index and locks it.

## One extra, easy to drop

Writing the demo for the narrowed matching exposed that `atlas list --category <unmatched>` printed "The atlas index publishes no entries" — false, and it sends an agent off to author a corpus the atlas may already have under a differently-spelled category. Narrowing the filter makes that reachable more often, so I fixed it in 6f502a0, kept as a separate commit on top so it can be dropped without disturbing the review round.

Also took your `--source` note as confirmation and made the private-corpus case explicit in the usage text and docs, since it's the answer to a question a reader will have:

```
sightmap atlas add toast-pos --source https://internal.corp/{slug}.tar.gz
```

`client.go`, `validate.go`, and `cache.go` are byte-identical to the previous revision, so the transport policy, escaping, and caps are untouched. Full suite and `-race` green; three-verb demo re-run against a local atlas.

jurassix (approved):
lgtm - full e2e proof doc attached https://app.staging.fullstory.com/subtext/o-B4XX-na1/docs/zWmpXZDBYBVa

go/atlas/atlas.go
```diff
@@ -0,0 +1,330 @@
...
+	Name         string   `json:"name,omitempty"`
+	Description  string   `json:"description,omitempty"`
+	Domains      []string `json:"domains,omitempty"`
+	Categories   []string `json:"categories,omitempty"`
+	Stats        Stats    `json:"stats"`
+	LastVerified string   `json:"last_verified,omitempty"`
+}
+
+// Stats is how much of a site an entry maps.
+type Stats struct {
```
joelgwebber: I'm a little hazy on why we're creating a new `type Stats`, as opposed to reusing the more complete version from #167. I worry this risks drift across the two implementations as we evolve them.

joelgwebber (approved):
Approving for when you're ready -- I'd just like to address the `Stats` duplication against #167.

joelgwebber:
Groovy, new approach lgtm!

Issue #157, opened 2026-08-03
SEP-0005: accept request property extraction (properties[]) into the schema


jurassix:
`properties: RequestProperty[]` lets a `Request` entry declare a named value extracted from a live request's own body or headers (`field` for a JSON dot-path, `pattern` for a regex against text) — the request-side analogue of SEP-0003's DOM `properties[]`. It answers what a `200 OK` response's body actually says (e.g. `{"status": "declined"}`), which no existing structured field (method, route, status code) can express.

This issue tracks accepting [SEP-0005](https://github.com/sightmap/sightmap/blob/main/spec/seps/0005-request-properties.md). Implementation: #112. Stays open until the SEP is accepted and the implementation lands.

joelgwebber:
Good overall approach. Just a couple of questions:

In an example like `pattern: 'rsp\.headers\.X-RateLimit-Remaining:\s*(\d+)'`, if I'm following correctly, this would allow the author to express *any* regex across things like `rsp.headers.X-Whatevs`, `rsp.body....`, etc, right? This could be quite powerful, but it could also be something of a foot-gun to know that you *must* write patterns to include those meta-fields. Another approach to consider:
```
- name: CheckoutPayment
  route: /api/checkout/pay
  method: POST
  properties:
    - name: status
      source: rsp.body
      field: .status  # maybe simplified jq-ish syntax, as in open-questions?
```

Then for the pattern case, we'd have something like:
```
properties:
  - name: rate_limit_remaining
    source: rsp.headers
    field: X-RateLimit-Remaining
    pattern: '(\d+)'  # Regex not really necessary in this case, but you get the idea.
```

Relatedly, I don't love the `transform:` stuff from SEP-0003 (my own damned fault; I vibed that out and never went back to fix it). The set of options are pretty dumb and arbitrary. But your idea here of just using regexes with a capture group would, I believe, generalize across all of those cases. So perhaps we could use that to eliminate the `transform:` altogether?

jurassix:
@joelgwebber the `source` addition looks good to me. Opening a pr with these suggestions to the sep.

jurassix:
@joelgwebber changes reflected in https://github.com/sightmap/sightmap/pull/171

joelgwebber:
Great, I think this update makes a big difference. I left an LGTM with a couple of small comments.

joelgwebber:
SEP-0005 is accepted and merged to `main`, and the implementation has landed:

[...]

PR #171, opened 2026-08-06
SEP-0005: source/field/pattern replaces oneOf(field, pattern)

jurassix:
Header extraction was only reachable through `pattern`, which had to encode both where to look and what to pull out in one regex. Joel's review on #157 called that a foot-gun and suggested splitting addressing (`source`/`field`) from extraction (`pattern`).

This rewrites SEP-0005 around that shape: `source` is a closed enum for the root, `field` selects a value within it, `pattern` optionally refines what `field` resolved (capture group 1, else full match) instead of being mutually exclusive with it. Headers now require `field`, so a raw regex can no longer smuggle in the address it's supposed to be scanning. Resolves the open questions on array indexing, capture groups, and pattern's missing root; adds two new ones on dotted JSON keys and whether this generalizes far enough to eventually retire SEP-0003's `transform` enum.

Prose-only change against `main`; the schema/implementation on #112 gets updated separately once this shape settles.

### DIFF
--- a/spec/seps/0005-request-properties.md
+++ b/spec/seps/0005-request-properties.md
@@ -4,7 +4,7 @@ title: Request property extraction via `properties[]`
 author: Clint Ayres (@jurassix)
 status: Draft
 created: 2026-07-31
-updated: 2026-08-06
+updated: 2026-08-10
 spec-version-target: 1
 related-issues: [157]
 related-discussions: []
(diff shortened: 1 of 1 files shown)


joelgwebber:
Looking good!

Small nit: Let's specify the regex dialect. The RE2 from Golang is a good choice, as (a) we'll use that, and (b) it's less strictly powerful, but more predictably performing than ECMA.

On the question of the `transform:` field, I agree we can leave it (and the undocumented `regex` value), addressing its cleanup and/or removal in a follow-up.

One useful addition to consider: There's currently no way to address URL paths/queries. Reasonable to defer, but worth considering.

joelgwebber (approved):
LGTM with one small nit and a follow-up.

PR #172, opened 2026-08-06
web: the atlas gallery — /atlas and /atlas/:slug

chiplay:
Builds the atlas gallery at `/atlas` and `/atlas/:slug`, and the machine twins the CLI reads: `/atlas/index.json`, `/atlas/<slug>.md`, and `/atlas/<slug>.tar.gz` for `sightmap atlas add`.


chiplay:
Pushed a copy pass over the detail view (71a45d5), from team feedback.

**In the entry README** (`web/src/data/atlas/sightmap-org.md`):

- Dropped the **"Why fewer than five views"** block. It answered a quality-bar question aimed at a reviewer, not at anyone reading about the site. The one fact worth keeping from it — the nav's Spec/Memory/Get started links are in-page anchors, not routes — moved into Coverage as a plain statement.
- **"What's worth knowing" → "Notes."** The section is a list of operational gotchas, so the plainest accurate word wins. Its self-referential preamble ("the notes are the point of a sightmap, so the interesting ones up front") went with it. Flagging the wording choice since the suggestion was "Summary" — the content isn't a summary, it's caveats, so "Notes" seemed the more honest label. Easy to change if you'd rather.
- **"Known gaps"** kept its name and content.
- Tightened throughout: fewer em dashes, and concrete behaviour in place of "will disagree with itself" / "took a wrong turn" (the last one now just names 200 as the status an unmatched path returns).

**In the page chrome:**

- The `/atlas` description no longer restates the heading directly above it.
- `1 of 1 entry` was noise; the count now reads `1 entry` until a filter is actually hiding something.
- Dropped the metadata rail's `Files` row — a bare count of vendored paths told a reader nothing.

### ⚠️ This edits a file marked do-not-hand-edit

`web/src/data/atlas/` carries its own README: *"Do not hand-edit. This directory is written by automation in the sightmap/atlas repo."* The entry was hand-vendored to unblock this phase, so editing it here is what makes the site correct today — but **the same edit has to land on `entries/sightmap-org/README.md` in `sightmap/atlas`, or the next `publish.yml` bot PR reverts it.** That repo isn't in this session's scope; happy to open the matching PR there if someone adds it.

`pnpm build` and `pnpm test` (100 passing) still green, and the browser checks re-ran clean — no horizontal scroll at 1280×900 or 390×844, single `<h1>`, no third-party requests.

---
_Generated by [Claude Code](https://claude.ai/code)_

web/src/lib/atlas.ts
```diff
@@ -0,0 +1,82 @@
...
+}
+
+/**
+ * Case-insensitive substring match across the fields a visitor would plausibly
+ * type: name, domains, description, categories, author.
+ */
+export function matchesQuery(entry: AtlasEntry, query: string): boolean {
+  const q = query.trim().toLowerCase()
+  if (!q) return true
+  const haystack = [entry.name, entry.description, entry.author, ...entry.domains, ...entry.categories]
```
jurassix: Add `entry.slug` to this list.

Thinking we add `sightmap atlas find` to the CLI reading this same index.json, and slug is what someone pastes back after seeing it on a card. Want the grid and the CLI matching the same fields so a query returns the same entries in both.

web/scripts/generate-feeds.ts
```diff
@@ -83,6 +113,60 @@ ${body}
...
+    `> ${oneLine(SITE_DESCRIPTION)}`,
+    '',
+    '## Docs',
+    '',
+    `- [Documentation](https://docs.sightmap.org): Guides, CLI reference, and the schema reference.`,
+    `- [Specification](https://github.com/sightmap/sightmap/tree/main/spec): The normative spec, JSON Schema, and conformance fixtures.`,
+    '',
+    '## Atlas',
+    '',
+    `> ${oneLine(ATLAS_DESCRIPTION)} Index: ${SITE_URL}/atlas/index.json`,
```
jurassix: Let's consider our approach here. This is where an agent that has never heard of us finds out the Atlas exists, so it should teach the lookup rather than just list entries.

The thing an agent has is a hostname. It knows `squareup.com`; it doesn't know the product is called Square POS or that the slug is `square-pos`. Right now it would have to guess.

Something like:

```
## Atlas

> Sightmaps of real websites: routes, components, selectors, network calls.
> Check here before exploring a site by hand.
>
>     sightmap atlas find <domain>    find a map by hostname
>     sightmap atlas add <slug>       install it into .sightmap/
>
> No CLI? Fetch https://sightmap.org/atlas/index.json and match your hostname
> against each entry's `domains[]`, then read the entry's .md twin below.

- [Sightmap](https://sightmap.org/atlas/sightmap-org.md): sightmap.org — Marketing site and blog for the Sightmap spec. 4 views, 28 components, 2 requests. Verified 2026-08-06. `sightmap atlas add sightmap-org`
```

Leading each line with the hostname means an agent can resolve straight from this file without fetching index.json at all.

`FeedAtlasEntry` would need three more fields to render that line: `domains`, `stats`, and `last_verified`. If that is more than you want here, the hostname alone carries most of the value and is the one I would not skip.

web/scripts/build-atlas.ts
```diff
@@ -0,0 +1,75 @@
...
+  const atlas = await loadAtlas(DATA_DIR)
+
+  fs.rmSync(PUBLIC_DIR, { recursive: true, force: true })
+  fs.mkdirSync(PUBLIC_DIR, { recursive: true })
+  fs.mkdirSync(GENERATED_DIR, { recursive: true })
+
+  // P4.3: /atlas/index.json, served byte-for-byte as vendored. Written from
+  // the original text rather than re-serialized from the parsed object so the
+  // route really is the file the atlas generator produced.
+  fs.writeFileSync(path.join(PUBLIC_DIR, 'index.json'), atlas.indexJson)
```
jurassix, line 45: Follow-up, not blocking.

The corpus isn't in the vendored tree. We get index.json, the .md, and screenshots, but not the six `.sightmap/*.yaml` files that index.json lists in `files[]`. So we can't serve `/atlas/<slug>.tar.gz`, which is what I'd like `add` to pull.

Two pieces: atlas CI vendors `<slug>/.sightmap/`, and this script writes the tarball next to index.json.

Rather do that than point `add` at raw.githubusercontent, for the reason you built the vendoring around. A takedown propagates to the gallery on rebuild, but if install goes to GitHub the entry stays installable and `removed.yaml` only binds the site.

web/src/components/atlas/AtlasInstall.tsx
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * The detail page's call to action: the one command that pulls this entry down
+ * locally, with a copy button.
+ *
+ * `sightmap add` ships in sightmap/sightmap#168, which was still in review when
+ * this page was built — the command is rendered anyway, on the handoff's
+ * instruction, because it is the page's whole point. If that PR lands under a
+ * different verb, this is the only place the string is written.
+ */
+export const installCommand = (slug: string): string => `sightmap add ${slug}`
```
jurassix: Heads up this may become `sightmap atlas add ${slug}`. I want to take #168 toward three verbs (`atlas list` / `atlas find` / `atlas add`), namespaced since `sightmap search` already means the local corpus. Still deciding on that vs a top-level `add`.

Nothing to change here, and keeping it in one function was the right call.

jurassix (commented):
Nothing to undo here. Two small things inline and one follow-up.

chiplay:
**Slug search** (2fc475a). `matchesQuery` includes `entry.slug`, so the grid and `sightmap atlas find` answer the same query with the same entries.

**llms.txt** (91f809c). `FeedAtlasEntry` gained `domains`, `stats`, and `last_verified`, and the section now teaches the lookup. As emitted:

```
## Atlas

> Community-contributed maps of views, components, and requests, made by browsing the live site with no source access. Check here before exploring a site by hand.
>
>     sightmap atlas find <domain>    find a map by hostname
>     sightmap atlas add <slug>       install it into .sightmap/
>
> Without the CLI, fetch https://sightmap.org/atlas/index.json and match your hostname against each entry's `domains[]`, then read that entry's .md twin below.

- [Sightmap](https://sightmap.org/atlas/sightmap-org.md): sightmap.org. Marketing site and blog for the Sightmap spec — landing page sections, blog index, and post pages. 4 views, 28 components, 2 requests. Verified 2026-08-06. `sightmap atlas add sightmap-org`
```

Three deliberate deviations from your sketch: all domains are listed rather than the first, so hostname matching works for multi-domain entries; an entry that lists none falls back to the `site_url` host, so the hostname is never missing; and the hostname is separated by a period rather than an em dash, because the vendored description already contains one.

**Install command** (6bf52d0). Now `sightmap atlas add ${slug}`, structure unchanged. The only other occurrence in the tree was `web/.sightmap/app.yaml`'s description, also updated. `HANDOFF-phase-4.md` still says `sightmap add`, left alone on purpose since it records what was asked at handoff time and says so, but happy to correct it if you'd rather it not read as current.

**The tarball** (43420df). Corpus vendored to `web/src/data/atlas/sightmap-org/.sightmap/` (all six blob SHAs match `seed/sightmap-org` exactly), and `build-atlas.ts` writes `public/atlas/<slug>.tar.gz`, gitignored so it stays build output.

No new dependency. `web/scripts/lib/tar.ts` is an ~80-line ustar writer, because package.json had no tar library and a general-purpose one would need configuring *away* from emitting exactly the member types the CLI refuses. Mode 0644 and mtime pinned to 0, so builds are byte-reproducible. `resolveCorpus()` screens for what the extractor rejects (symlinks, non-regular files, unsafe or over-long names) and its caps (4 MiB per file, 32 MiB total, 512 members); a corpus that trips any of it goes unpublished with a warning rather than failing the build, matching how a bad entry is already handled.

End-to-end against the real CLI, `dist/` served on loopback:

```
$ sightmap atlas add sightmap-org --source http://127.0.0.1:8792/atlas/{slug}.tar.gz
  wrote  .sightmap/blog-post.yaml
  ... 6 files ...
Installed sightmap-org: 6 files → .sightmap. Next:
  sightmap validate

$ sightmap validate
✓ no validation errors
```

`--source` was needed only because that CLI build predated the re-point. #168 now defaults to `https://sightmap.org/atlas/{slug}.tar.gz`, which resolves to `/atlas/sightmap-org.tar.gz` — exactly what this build emits, so the bare command hits this artifact once both land. Server log shows the only fetches were the index and the tarball.

`pnpm build` green, `pnpm test` 113 passing across 9 files.

**One thing worth a separate decision:** `.github/workflows/web.yml` runs `pnpm build` but never `pnpm test`, so the suite — including the three test files added here — doesn't run on a PR. Out of scope for this branch, but it means the coverage we keep adding isn't gating anything.

Issue #189, opened 2026-08-11
Restructure the reference Go library for downstream consumers


joelgwebber:
## Summary

`github.com/sightmap/sightmap/go` is the shared, spec-backed implementation that
external consumers build on. Before it has real consumers, restructure it into a
clean, layered surface so that (a) all corpus / observed / result types come from
one obvious package, (b) the matching engines are consistent and discoverable,
and (c) reusing the model + matching never drags in the CDP browser tool.

Sibling of #160 (Corpus/Matcher split, `Def` naming, wire IR) — this generalizes
that work across the whole library surface.

## Motivation

The surface is split and partly conflated in ways that will ossify once consumed:

- **Types split across packages.** `ComponentDef` / `Property` live in the
  `match` engine package; the rest in `sightmap`. Even the "clean" façade leaks
  `match` and `comps` into consumer signatures.
- **Redundant engine front doors.** Three public paths into the component NFA
  (`Matcher.MatchTree`, `ApplySightmap`, `FindConflicts`), and a
  `sightmap.Matcher` type that collides conceptually with the `match` package.
- **Observed types trapped in the CDP tool.** Network/console records live inside
  `browser` and are entangled with a live CDP connection (`NetworkEntry.conn`),
  so any consumer with its own capture must depend on the browser driver to reuse
  matching/normalization.
- **`SelectorPart` double-duty.** One struct serves as both an observed element's
  identity and a matching pattern, carrying predicate-only fields (`AttrOps`,
  `:not`/`:is`/`:has`) that are meaningless on an observed node.

## Target shape

```
sightmap   the MODEL: all vocabulary (corpus defs, observed records, match
           results, property outputs) + loader + validate + the atomic selector
           parse/match primitive. Self-contained leaf (absorbs comps + sel).
match      the ENGINES: the composed component NFA + request/message/view
           matchers.                                             [-> sightmap]
browser    the CDP tool (name kept); extraction/normalization made reusable
           WITHOUT the driver.                                   [-> sightmap, match]
render     formatting.            cmd/sightmap   CLI.
```

Consumer story: **one import for types (`sightmap`), one for engines (`match`)**;
a consumer with its own capture never imports the CDP tool.

## Naming model

- `...Def` = a spec matched/checked against an observed instance: `ComponentDef`,
  `RequestDef`, `MessageDef`, `ViewDef`.
- Observed records are bare (`Element`, `Request`, `Message`); results are
[...]

PR #190, opened 2026-08-11
sightmap: split observed Element identity from the SelectorPart pattern

joelgwebber:
First structural slice of the library restructuring for downstream consumers
(#189): split the observed **element identity** from the **selector pattern**.

`ComponentNode.Selector` was doing double duty — the observed identity of a DOM
element (its `tag`/`id`/`classes`/`attrs`) *and* a matching pattern, carrying
predicate-only fields (`AttrOps`/`Not`/`Is`/`Has`) that are meaningless on a real
observed element. This introduces `comps.Element` for the observed *subject* and
matches it against `SelectorPart` *patterns*, so the matcher signature is honest
(`Matches(el *Element, rule *SelectorPart)`) and an observed element structurally
can't hold a `:not()`.

## Changes

- **`comps`**: add `Element{Tag,Id,Classes,Attrs}`; `ComponentNode.Selector`
  (`*SelectorPart`, json `selector`) → `Element` (`*Element`, json `element`).
- **`sel`**: `Matches(el *Element, rule *SelectorPart)`; `effectiveAttrValue`
  over `Element`; `MatchesNode` reads `node.Element`.
- **`extract`**: `elementFor` projects the parsed compound into an `Element`.
- **`render` / `coverage` / `browser` / `cmd`**: read `node.Element`;
  observed-side display helpers took a type change only (bodies unchanged —
  `Element` is the same `tag/id/classes/attrs` subset).
- **fixtures + `component-schema.md`**: the structured `"selector"` key on
  component-tree JSON renamed to `"element"`.

## Notes

- **Behavior-neutral.** The observed side simply drops predicate fields it never
  populated. `ProbeComponent.Selector` (a *string*, a different struct) is
  untouched, so `probe.js` needs no change.
- **Verification:** `go build`/`vet`/`test ./...`, plus the `spec` conformance
  and examples validators, all pass; gofmt clean.
- A follow-up (untangling the observed network/console records + extraction from
  the CDP `browser` tool) will **stack on top of this branch**.

### DIFF
--- a/go/coverage/anchor.go
+++ b/go/coverage/anchor.go
@@ -17,8 +17,8 @@ import (
 func NearestDataAttrAncestor(node *comps.ComponentNode, parentMap ParentMap) *comps.ComponentNode {
 	curr := parentMap[node]
 	for curr != nil {
-		if curr.Selector != nil {
-			a := curr.Selector.Attrs
+		if curr.Element != nil {
+			a := curr.Element.Attrs
 			if a["data-testid"] != "" || a["data-component"] != "" {
 				return curr
 			}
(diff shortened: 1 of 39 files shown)


nrodd (approved):
Looks good to me!

go/coverage/anchor.go
```diff
@@ -82,22 +82,22 @@ func ArrowHint(node *comps.ComponentNode, parentMap ParentMap) string {
...
-	return SelectorString(anc.Selector) + " " + nodeTag
+	return SelectorString(anc.Element) + " " + nodeTag
 }
 
-// SelectorString formats a SelectorPart as tag[#id][.cls1.cls2]. Attributes are
-// not included; they appear only in the trace/anchor displays above.
-func SelectorString(s *comps.SelectorPart) string {
+// SelectorString formats an observed Element as tag[#id][.cls1.cls2]. Attributes
+// are not included; they appear only in the trace/anchor displays above.
+func SelectorString(s *comps.Element) string {
```
sirrah23, line 100: Optional minor nit: it looks like we have this function in two places -- here & in `go/browser/actions.go` We might benefit from a util package that has this in it. Alternatively `*comps.Element` could expose a method for this. Or the `comps` package itself could expose a helper. Food for thought.

sirrah23 (approved):
One minor nit but o/w LGTM!

go/coverage/anchor.go
joelgwebber: Thanks, good catch. I'll fix forward in the next PR in the stack.

PR #191, opened 2026-08-11
sightmap: lift observed Request/Message records out of the browser tool

joelgwebber:
Second structural slice of the library restructuring for downstream consumers
(#189). **Stacked on #190** (base branch `refactor/observed-element-identity`);
review/merge that first.

Untangles the observed network/console records from the CDP `browser` tool. They
lived inside the tool and — for network — were entangled with a live CDP
connection, so a consumer with its own capture (e.g. a session-review pipeline)
had to depend on the browser driver just to reuse the records. This moves the
pure records into the `sightmap` model as the observed counterparts of
`RequestDef` / `MessageDef`.

## Changes

- **`sightmap`**: add observed records `Request` and `Message` — tool-free,
  wire-friendly, carrying no connection or fetch handle.
- **`browser`**: `ConsoleEntry` → `sightmap.Message` (it was already pure).
  `NetworkEntry` becomes an internal `networkRecord` that embeds
  `sightmap.Request` plus the **unexported** CDP fetch handle
  (`requestID` + `conn`). `Collector.Console`/`Network` now return the pure
  `sightmap` records; lazy response-body fetch keeps using the handle. `browser`
  now imports `sightmap` (acyclic — `sightmap` never imports `browser`).
- **`cmd/sightmap` devtools**: consume `sightmap.Message` / `sightmap.Request`.

## Notes

- **Behavior-neutral.** The public devtools JSON is unchanged: the embedded
  `Request`'s fields promote inline in JSON, and the handle stays unexported.
- `extract` (tree normalization) already has no `browser` dependency, so a
  consumer can now produce a `ComponentNode` tree + `sightmap` records and match
  them without importing the CDP driver at all.
- **Verification:** `go build`/`vet`/`test ./...`, plus `spec` conformance and
  examples validators, all pass; gofmt clean.

---

**Update (rebased onto `main` after #190 merged).** Also folds in a small
opportunistic cleanup: `coverage.SelectorString` and `browser.selectorString`
were byte-identical `tag[#id][.cls]` formatters over `*comps.Element` — merged
into one nil-safe method, `comps.(*Element).SelectorString`, with both call sites
repointed. Separate commit in this PR.

### DIFF
--- a/go/comps/component.go
+++ b/go/comps/component.go
@@ -1,6 +1,9 @@
 package comps
 
-import "encoding/json"
+import (
+	"encoding/json"
+	"strings"
+)
 
 // SelectorPart is the structured CSS identity of a DOM element, or a
 // synthetic identity for native mobile elements. It carries no proto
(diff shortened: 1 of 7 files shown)


go/comps/component.go
```diff
@@ -58,6 +61,26 @@ type Element struct {
 	Attrs   map[string]string `json:"attrs,omitempty"`
 }
 
+// SelectorString formats the element as a simple CSS selector, tag[#id][.cls...].
```
joelgwebber, line 64: De-dupe leftover from #190.

go/browser/tmp.yaml
nrodd, line 1: Should this get removed and added to the gitignore ?

nrodd (approved):
LGTM, just had one potential nit?

go/browser/tmp.yaml
joelgwebber: D'oh, good catch. Will remove!

PR #192, opened 2026-08-11
sightmap: move component types to the model; flip the sightmap↔match edge

joelgwebber:
First consolidation step of the library restructuring (#189). **Stacked on #191**
(base `refactor/observed-runtime-records`); review/merge that first.

Makes `sightmap` the home for the component data types and **inverts the
dependency** so the engine (`match`) depends on the model, not the reverse — the
change that lets a consumer import `sightmap` for corpus types without pulling in
the matcher.

## Changes

- **`sightmap`**: `ComponentDef`, `Property`, `ComponentMatch`, `Conflict` move
  here from `match` (new `component.go`). **Production `sightmap` no longer
  imports `match`.**
- **`match`**: keeps the NFA engine (`ParseQueries` / `ApplySightmap` /
  `FindConflicts` / `FindAllMatches`) and now consumes the `sightmap` types. The
  corpus `Matcher` wrapper moves here from `sightmap` (`corpus_matcher.go`) as
  `match.Matcher` / `match.NewMatcher`.
- **callers**: `match.ComponentDef` → `sightmap.ComponentDef` (etc.);
  `sightmap.NewMatcher` → `match.NewMatcher` across
  `cmd`/`observe`/`render`/`coverage`/`viewset`/`compquery`.

## Note on scope

Pure **relocation + edge flip** — no renames (`View`/`Property` keep their names)
and `comps`/`sel` are not yet folded into `sightmap`; those are follow-up slices.

The `Matcher`-wrapper move (nominally the next step's territory) **had to ride
with this change**: moving `ComponentDef` into `sightmap` forces `match` to
import `sightmap`, so the wrapper — which calls into the NFA — could not stay in
`sightmap` without creating a `sightmap ↔ match` cycle. What remains for the
"engines" step is the API cleanup (collapse the redundant `MatchTree` /
`ApplySightmap` / `FindConflicts` front doors into one `Matcher`), not the move
itself.

## Verification

`go build`/`vet`/`test ./...`, `spec` conformance + examples, gofmt — all green.
Behavior unchanged.

### DIFF
--- a/go/browser/actions.go
+++ b/go/browser/actions.go
@@ -448,21 +448,6 @@ func ScreenshotWithOptions(ctx context.Context, conn *CDPConn, opts ScreenshotOp
 	return data, nil
 }
 
-// selectorString builds a simple CSS selector string from an observed Element.
-func selectorString(s *comps.Element) string {
-	if s == nil {
-		return ""
-	}
-	b := s.Tag
-	if s.Id != "" {
-		b += "#" + s.Id
-	}
-	for _, c := range s.Classes {
-		b += "." + c
-	}
-	return b
-}
-
 // ResolveBySightmapID resolves a probe id to a live element via its persisted
 // data-sightmap-id attribute (set on the DOM during the prior extraction) and
 // returns a ComponentNode carrying the element's CURRENT bounding box.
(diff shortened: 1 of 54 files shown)


joelgwebber:
Folded into #193 — these two changes touch 63% of the same files, and the combined diff is ~the same size, so reviewing them as one PR (kept as two commits: edge-flip, then renames) is clearer. #193 now contains both commits stacked on #191.

PR #193, opened 2026-08-11
sightmap: consolidate component types into the model + apply the Def naming convention

joelgwebber:
Consolidation step of the library restructuring (#189): make `sightmap` the home
for the component model, and apply the `Def` naming convention. Based on `main`
(#190/#191 merged).

Two commits — review per-commit:

**1. Move component types to the model; flip the `sightmap ↔ match` edge**
- `ComponentDef`, `Property`, `ComponentMatch`, `Conflict` move `match` →
  `sightmap`. Production `sightmap` no longer imports `match`.
- `match` keeps the NFA engine and now depends on `sightmap`; the corpus
  `Matcher` wrapper moves to `match` (it had to ride along — moving `ComponentDef`
  into `sightmap` forces `match → sightmap`, so the wrapper couldn't stay in
  `sightmap` without a cycle).

**2. Rename to the `Def` naming model**
- `View` → `ViewDef`, `Property` → `ComponentPropertyDef`,
  `RequestProperty` → `RequestPropertyDef`.
- Pure identifier rename — **JSON tags unchanged, wire byte-identical**. `View`
  used `gorename` (semantic) so the user-facing `"[View: NAME]"` snap header and
  CLI column labels are untouched; property renames are word-boundary text
  renames (no string-literal hits).

_(Originally split as #192 + #193; merged here since they touch 63% of the same
files and the combined diff is ~the same size. Kept as two commits so the
architectural edge-flip stays reviewable on its own.)_

## Verification
`go build`/`vet`/`test ./...`, `spec` conformance + examples, gofmt — all green.
Behavior unchanged. `ViewMatch` intentionally not added (no producer yet).


go/sightmap/corpus.go
```diff
@@ -76,7 +74,7 @@ type View struct {
 }
 
 // ViewByName returns a pointer to the first View with the given name, or nil.
```
nrodd, line 76:
```suggestion
// ViewByName returns a pointer to the first ViewDef with the given name, or nil.
```

nrodd (approved):
LGTM, just one small nit

go/sightmap/corpus.go
joelgwebber: Got it, thanks!

Issue #170, opened 2026-08-06
Runtime matching for messages: and request properties: (evaluation half of SEP-0005/0006)


joelgwebber:
## Summary

The runtime/evaluation half of the SEP-0005 (`requests:` `properties:`) and
SEP-0006 (`messages:`) work. Both entities landed as **declaration + validation
only** in `go/sightmap`; nothing yet derives live records from a session and
matches/extracts against the definitions. This issue tracks both runtime halves
together — they share the same shape (a live record type + a corpus-aware
predicate alongside the def) and the same precedent (`Corpus.RequestsForURL`
from #47).

Framed at the **library boundary**: the deliverable is the lib predicate a
downstream consumer (FullStory/lidar, internal SUBTEXT-480; see #160) calls
directly — not enrichment of the `sightmap` CLI devtools UI. (#47 itself closed
at that boundary: `RequestsForURL` landed with no CLI consumer, which is fine.)

### A. Console/exception messages (`messages:`, SEP-0006)

1. **Record type + CDP derivation — largely already exists.** `go/browser/collector.go`
   already captures console + exceptions into `ConsoleEntry{Level, Text, Ts,
   Index, Tab}`, folds uncaught exceptions in as `level: "exception"` (the exact
   SEP-0006 correction), and normalizes CDP `warning` → `warn` via `consoleLevel`.
   The six-level vocabulary (`log`/`debug`/`info`/`warn`/`error`/`exception`) is
   the empirical behavior SEP-0006 rests on. Consider lifting/sharing the record
   shape so the lib predicate doesn't depend on the `browser` package.
2. **The predicate + the deferred runtime MUST.** A `MessagesForRecord`-style
   function in `go/sightmap` alongside `MessageDef`: case-insensitive level
   equality + one RE2 `regexp` per def. Surface the multi-match ambiguity
   (SEP-0006 Conformance: *"MUST warn when a live record matches more than one
   entry"*), which the lib currently documents as a consumer obligation, not
   code. This does **not** belong in `go/match` (DOM selector / NFA machinery); a
   console record is a flat stream.

### B. Request-property extraction (`requests:` `properties:`, SEP-0005)

3. **A live req/rsp record type** — or reuse the collector's `NetworkEntry`,
   which already buffers request/response headers + bodies (confirm bodies are
   retained; cf. the lazy-`getResponseBody` limitation). Belongs in `go/sightmap`
   next to `RequestDef`/`RequestsForURL`.
[...]

PR #198, opened 2026-08-11
sightmap: runtime matching for console/exception messages (SEP-0006, #170)

joelgwebber:
Part of the runtime-matching work (#170), building on the c3a8 library restructuring (#189). Stacked on top of #196.

## What

Adds the runtime (family-B, flat) side of SEP-0006 messages: given an **observed** `Message`, find which `MessageDef`(s) in the corpus classify it.

- **`MessageMatch`** — the message-side analogue of `ComponentMatch`: a projection of the matched def (`Name`, `Description`, `Source`). Messages have no live-extracted properties, so it carries no property values.
- **`Corpus.MessagesForRecord(rec Message) []MessageMatch`** — returns every matching def in declaration order.

## Semantics (mirrors validation + the existing flat matchers)

- `Level` compared case-insensitively for equality; `Message` as an **RE2** regex (Go's `regexp`) against the record text; an omitted constraint matches anything (both omitted ⇒ matches every record).
- An uncaught exception arrives as level `exception`, so `level: error` defs don't claim it — covered by a test.
- Per-call `regexp.Compile`, matching the established `MatchRoute` idiom (no premature caching). A malformed pattern is **skipped, not panicked on**, so an unvalidated in-memory corpus stays silent; validation reports `message-regex-invalid` separately.

## Multi-match MUST

SEP-0006 requires a record matching more than one entry be surfaced, not silently resolved to a first match. Consistent with how `Matcher.Conflicts` returns every colliding claim, `MessagesForRecord` returns **all** matches with no winner — `len(result) > 1` is the ambiguity a consumer must surface.

## Not in scope

No wiring into a tool/CLI yet (nothing consumes `RequestsForURL` at runtime either); this lands the library primitive for consumers. Request-property extraction (SEP-0005 runtime) is the sibling task.

## Verification

`go build/vet/test ./...` green; `gofmt` clean; spec `validate:conformance` + `validate:examples` pass.

### DIFF
--- a/go/sightmap/message.go
+++ b/go/sightmap/message.go
@@ -1,5 +1,10 @@
 package sightmap
 
+import (
+	"regexp"
+	"strings"
+)
+
 // MessageDef is a named console-output or runtime-exception pattern from the
 // sightmap corpus (SEP-0006). It gives console activity the same thing
 // RequestDef gives network activity: a named entity other parts of the corpus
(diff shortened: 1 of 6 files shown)


go/sightmap/message.go
```diff
@@ -54,3 +59,52 @@ const (
...
+// separately (message-regex-invalid).
+func (c *Corpus) MessagesForRecord(rec Message) []MessageMatch {
+	var out []MessageMatch
+	for i := range c.Messages {
+		m := &c.Messages[i]
+		if m.Level != "" && !strings.EqualFold(m.Level, rec.Level) {
+			continue
+		}
+		if m.Message != "" {
+			re, err := regexp.Compile(m.Message)
```
sirrah23: Nit for consideration: instead of doing a `regexp.Compile` in this loop we could instead do it when we append the message to `c.Messages` and hold a reference to that regex on the message itself i.e. `m.Regex`. 

A minor thing but depending on how often this function is called and how many messages are the compiles _could_ add up over time.

sirrah23 (approved):
LGTM!

PR #199, opened 2026-08-12
cmd/sightmap: corpus-aware devtools — annotate console+network with matches (#170)

joelgwebber:
Part of the runtime-matching work (#170), stacked on top of #198.

## What

Makes the `browser start` daemon **corpus-aware** and gives the two flat matchers (`MessagesForRecord` from #198, and the already-landed `RequestsForURL`) their first non-test consumer. Until now the devtools path was corpus-blind: `sightmap console`/`network` rendered raw records with no corpus loaded anywhere.

The `/devtools/console` and `/devtools/network` handlers now annotate each observed record with the names of the corpus defs that classify it:

- **console** → `Corpus.MessagesForRecord` (level + message-regex)
- **network** → `Corpus.RequestsForURL` (route + method **identity**)

`console`/`network` `list` and `get` render the matched names.

## New Output
This winds request/message matches into the `sightmap browser` devtools output like so:

```
console:
[0] [CartVersionMismatch]  error     cart version mismatch
[1] [--]                   log       just a plain log line

network:
[1] [PageDocument]         GET  http://localhost:8799/                  → 200 OK (Document)
[2] [CheckoutPayment]      POST http://localhost:8799/api/checkout/pay  → 501 … (Fetch)
[5] [--]                   GET  http://localhost:8799/favicon.ico       → 404 … (Fetch)
```

## Design (per discussion)

- **Per-request corpus load.** The corpus is loaded fresh on each devtools request, so daemon annotations reflect on-disk edits exactly like the component CLI's per-invocation `sightmap.Load()`. Corpora are tiny; the I/O is negligible. (An mtime short-circuit is a trivial later optimization if needed.)
- **Handler-side, collector stays simple.** Annotation lives in the devtools handler; the collector remains a pure observation buffer.
- **Graceful degrade.** A missing or malformed corpus yields un-annotated entries rather than failing the query — devtools stays useful without a corpus.
- **Wire format.** Each entry gains a `"matches"` list; the record's own fields are promoted via embedding, so the shape is just the record plus `matches`.

## Relation to the daemon north star (#85a4-ish)

This is a deliberate first corpus-aware step for the daemon, not throwaway: the north star (route all browser ops through the daemon, CLI/MCP as thin front-ends) explicitly grows the protocol *from* the `/devtools/*` query surface, and a corpus-aware daemon returning matched records is exactly that curated-op shape.

## Not in scope

- **Request property/signal extraction** — needs req/rsp headers + bodies (the buffered record has neither; bodies are lazily fetched), so it lands separately and layers onto this same handler annotation.
- **Eliminating the double corpus load** (component CLI still loads client-side) — a later daemon-convergence step.

## Verification

`go build/vet/test ./...` green; `gofmt` clean; spec `validate:conformance` + `validate:examples` pass. Pure annotation helpers (`annotateConsole`/`annotateNetwork`) are unit-tested, including nil-corpus degrade.

### DIFF
--- a/go/cmd/sightmap/cmd_browser_start.go
+++ b/go/cmd/sightmap/cmd_browser_start.go
@@ -153,7 +153,7 @@ func runBrowserStart(args []string) error {
 	// Devtools query surface. The collector is created after Chrome is ready
 	// (below); the handlers return 503 until then.
 	var collectorPtr atomic.Pointer[browser.Collector]
-	registerDevtoolsHandlers(mux, &collectorPtr)
+	registerDevtoolsHandlers(mux, &collectorPtr, *sightmapDir)
 
 	// Bind the sightmap server on the IPv4 loopback (not the ":port" wildcard) so
 	// it shares an address family with Chrome's CDP and with FindFreePort's probe.
--- a/go/cmd/sightmap/cmd_devtools.go
+++ b/go/cmd/sightmap/cmd_devtools.go
@@ -51,14 +51,70 @@ func devtoolsGet(sightmapDir, path string, query url.Values) ([]byte, error) {
 	return body, nil
 }
 
+// consoleEntry / networkEntry are an observed record plus the names of the
+// corpus definitions that classify it. The daemon annotates entries in the
+// devtools handler (the collector stays corpus-blind); Matches is empty when no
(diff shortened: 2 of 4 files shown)


sirrah23 (approved):
LGTM!

Issue #160, opened 2026-08-03
Library architecture: canonical Corpus (data) vs Matcher (engine), Def naming, and a flat corpus wire IR


joelgwebber:
## Summary

Formalize the runtime data model so the library is a clean dependency for downstream consumers that (a) match against live/observed trees to spec and (b) ship a whole corpus over a wire without a lossy re-compile.

Motivated by the FullStory/lidar consumer (internal SUBTEXT-480), which is moving its session-review matching off a duplicate in-repo matcher onto this library. Two friction points today:

- `sightmap.Corpus` / `View` carry **no JSON tags**, so there is no blessed wire/storage form; `cmd/sightmap` serve hand-rolls a lossy `compiledSightmapJSON` that drops information and is CLI-local.
- The corpus's **structure** (view routes, view-scoped gating, requests) has no compact serialized form the consumer can reconstruct and match per-URL.

None of this changes matching semantics; it is a naming + layering + serialization cleanup, plus one new query surface (`RequestsForURL`).

## Proposed model

**Data vs engine split**
- `Corpus` = pure, JSON-tagged **data** — the wire, storage, and serve payload, all one representation:
  ```
  Corpus{ memory[], globals[]ComponentDef, requests[]RequestDef, views[]View }
  View{ name, route, memory[], components[]ComponentDef, requests[]RequestDef }
  ```
- `Matcher` = **engine** built from a `*Corpus` (`NewMatcher(*Corpus)`), holding the compiled NFA cache; never serialized. This deletes serve's `compiledSightmapJSON`.

**Naming — the `Def` convention**
- Rule/authoring side gets a `Def` suffix; the observed runtime construct keeps the bare noun:
  - a `ComponentDef` *matches* a `Component`; a `RequestDef` *classifies* a `Request`.
  - `Corpus` = data, `Matcher` = engine, `Match` = result.
- The observed `Component`/`Request` node types are a separate, partly-shared concern (see `comps.ComponentNode`); not this issue's focus.

**Flat wire IR (source to IR split)**
- Authoring types (`raw*`: hierarchical, `$ref`, string-or-list selectors) stay as the parse layer and **compile into a flat IR**.
- The flat `ComponentDef` carries:
  - `selectors[]` — compound selectors, the **only** matcher input.
  - `parentChain[]` — ancestor component **names** (root-first), for classification/display only (lint/coverage/observe/serve). Never matched on; `omitempty`.
[...]

PR #200, opened 2026-08-12
cmd/sightmap: serve the canonical Corpus wire (retire compiledSightmapJSON) (#160)

joelgwebber:
Closes the remaining substantive piece of #160 (fbba.7).

## What

The sightmap server — both `serve-sightmap` and the `browser start` daemon, which share `compileCorpus` + the `/sightmap` handler — emitted a bespoke `compiledSightmapJSON`: a lossy re-compile of the corpus into the extension's private shape (joined-selector strings, a `viewComponents` route→components map, a flat `globals` list).

This replaces it with a thin transport envelope `{site, version, corpus}`, where **`corpus` is the canonical `Corpus` wire** — `selectors: []` arrays, components nested under each view, plus `requests`/`messages` — byte-identical to what a `go get` / library consumer (subtext, lidar) sees.

## Why

- **The browser extension becomes an internal consumer of the same wire we ship downstream.** That dogfoods the format against a real consumer *before* external parties lock in, and it's the cheapest place to surface any wire gaps.
- **Deletes an entire representation + a compile step** (`compiledComponent/Prop/View/SightmapJSON`, `normaliseComponent`) — net entropy reduction.

## Shape

`GET /sightmap`:
```json
{ "site": "smtest", "version": "1786…",
  "corpus": { "globals": [ { "name": "PageHeading", "selectors": ["h1"] } ],
              "views":   [ { "name": "Home", "route": "/",
                             "components": [ { "name": "TestBody", "selectors": ["body"] } ] } ],
              "requests": [ … ], "messages": [ … ] } }
```

- The `corpus` body is the pure canonical wire; `site`/`version` are serve-only transport metadata in the envelope.
- **Cache-busting unchanged:** `/sightmap/version` still serves the stamp; the extension polls it and reloads on change. (Kept the version in the envelope body rather than an ETag header because the extension fetches through a background service worker — reading response headers there would need service-worker + CORS-expose changes for no wire-purity gain, since `corpus` is already the bare canonical wire.)

## Extension

The selector-join and view flattening move client-side (`content.js` `normalizeComp`: `selectors[].join(", ")` → a `closest()`-compatible string), so **`resolver.js` is untouched**. `background.js`'s component-count helper reads the new nested shape. Server + extension change together in one PR (the extension is `go:embed`ed into the binary), so there's no half-broken intermediate.

## Verification

`go build/vet/test ./...` green, `gofmt` clean. Verified **end-to-end** against a live headless session (`--extensions` pointed at the repo dir): the extension logged `loaded 2 components` — a global + a routed view component — parsed from the new `selectors[]` wire, and the served JSON has the shape above.

### DIFF
--- a/.changeset/serve-corpus-wire.md
+++ b/.changeset/serve-corpus-wire.md
@@ -1,0 +1,9 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`sightmap serve` and the `browser start` daemon now serve the canonical Corpus
+wire — the same shape library consumers see (`selectors[]` arrays, components
+nested under each view) — inside a thin `{site, version, corpus}` envelope,
+replacing the bespoke pre-compiled shape. The bundled browser extension consumes
+it directly; cache-busting via `/sightmap/version` is unchanged.
--- a/go/cmd/sightmap/cmd_browser_start.go
+++ b/go/cmd/sightmap/cmd_browser_start.go
@@ -109,28 +109,28 @@ func runBrowserStart(args []string) error {
 
 	// ── Start sightmap HTTP server in background ──────────────────────────────
 	siteName := filepath.Base(cwd())
-	compiled, err := compileCorpus(*sightmapDir, siteName)
+	compiled, err := loadServedSightmap(*sightmapDir, siteName)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


sirrah23 (approved):
LGTM!

Issue #206, opened 2026-08-13
`messages:` can't match or extract on exception stack traces


joelgwebber:
`MessageDef` matches on `level` + a regex over the record's `text`, and the observed record is `Message{Index, Tab, Level, Text, Ts}`. An uncaught exception's **stack** — its throwing origin and frames — is dropped, so a corpus can't classify an exception by where it came from, nor extract the failing location.

## Proposed record extension (additive)

```go
type Message struct {
    // …existing…
    Stack []Frame   // empty for non-exceptions
}
type Frame struct { Function, File string; Line, Column int }
```

## Def side (small SEP-0006 follow-on)

Give `MessageDef` a stack-addressing `properties[]` mirroring SEP-0005's `source` / `field` / `pattern` — e.g. `source: stack`, `field: top.file` or a frame index, optional `pattern` — so an exception can be named/classified by origin and its location extracted.

## Relationships

- Extends the message half of #170 (matching is `level` + `text` only today) and SEP-0006 (#113).
- Capture side: the collector already receives CDP `Runtime.exceptionThrown` `stackTrace.callFrames` but drops them.
- The message analogue of the request-body gap — same "records faithful to defs" principle.

Issue #130, opened 2026-07-31
Network collector: WebSocket traffic invisible (capture WS upgrades + optional frame logging)


joelgwebber:
The collector captures HTTP request/response only; `network list --url ws`
returns nothing, so a WebSocket-driven app shows every asset and beacon *except*
its actual traffic, and a `requests:` entry for a WS endpoint can never match.

**Expected:** capture the WS upgrade request at minimum (ideally with optional
frame logging).

Split from the original omnibus (2026-08-16): the HTTP response header/body
retention + surfacing gaps moved to #226 (needed to wrap up the request-property
runtime work). This issue is WebSocket visibility only — the larger, separable
effort.

PR #207, opened 2026-08-13
sightmap: faithful Request records + RequestsForRecord property extraction (SEP-0005 runtime half)

### LINKED ISSUE #205: Observed `Request` records carry no headers or bodies, so SEP-0005 `properties:` can't be evaluated
SEP-0005 (landed, #157 / #112) lets a `RequestDef` declare `properties[]` whose `source` is one of `req.body` / `rsp.body` / `req.headers` / `rsp.headers`. But the observed record it must run against —

```go
Request{Index, Tab, Method, URL, Status, StatusText, ResourceType, Ts}
```

carries **neither bodies nor headers**. The def surface has outgrown the record: a corpus can say "extract `declineReason` from `rsp.body`," and nothing in the runtime record can supply it.

## Principle

An observed record should faithfully carry everything its matching def can address — kept **optional** so a lazy/live producer may omit expensive fields (fetch-on-demand) and matching degrades to "property absent," while a producer holding a full capture populates them.

## Proposed record extension (additive, all optional)

```go
type Request struct {
    // …existing…
    DurationMs int64
    ReqHeaders []Header   // ordered, duplicate-preserving; case-insensitive lookup
    RspHeaders []Header
    ReqBody    *Body
    RspBody    *Body
}
type Header struct { Name, Value string }
type Body struct { Content string; Size int; ContentType string; Truncated bool }
```

- Raw `Content` + `ContentType`; the **matcher** parses (the `field` dot-path implies JSON at match time — keep parsing out of the record).
- `Truncated` / `Size` let a matcher distinguish "absent" from "absent because truncated."
- Add a record-based **`RequestsForRecord(rec) []RequestMatch`** that does identity match **and** resolves `properties[]` — the SEP-0005 runtime half.

## Relationships

- Record-shape prerequisite for the request-extraction half of #170 (which currently proposes reusing the collector's `NetworkEntry`).
- Capture side is #130 (response headers/bodies not retained).
- Resolves the lean-vs-faithful tension the current record comment names ("bodies are large") by making the fields optional rather than absent.

joelgwebber:
Closes #205. The runtime (evaluation) half of SEP-0005 request `properties:`, at the library boundary — the surface a downstream consumer (FullStory/lidar) calls directly.

## Problem

A `RequestDef` can declare `properties[]` whose `source` is `req.body` / `rsp.body` / `req.headers` / `rsp.headers`, but the observed record it must run against —

```go
Request{Index, Tab, Method, URL, Status, StatusText, ResourceType, Ts}
```

carried neither bodies nor headers. The def surface had outgrown the record: a corpus could say "extract `outcome` from `rsp.body`," and nothing in the runtime record could supply it.

## Change

**Record (`records.go`), additive + all optional** — faithful to what a def can address, kept optional so a lazy/live producer may omit expensive fields and matching degrades to "property absent," while a full-capture producer populates them:

```go
Request{ …existing…
    DurationMs int64       // the reserved "duration" identity
    ReqHeaders []Header     // ordered, duplicate-preserving; case-insensitive lookup
    RspHeaders []Header
    ReqBody    *Body        // raw Content + ContentType; matcher parses
    RspBody    *Body
}
Header{ Name, Value string }
Body{ Content string; Size int; ContentType string; Truncated bool }
```

All `omitempty`, so an unpopulated record marshals byte-identically to before (regression-guarded by a test).

**Matcher (`request_extract.go`)** — `Corpus.RequestsForRecord(rec) []RequestMatch`: route+method identity via the existing `RequestsForURL` (one identity path), then per matched def resolves `properties[]` into a `RequestMatch{Name, Memory, Tags, Properties}` (the request-side analogue of `ComponentMatch`). Resolution is `source → field → pattern → transform`:

- `field` on a `*.body` walks a JSON dot-path (object keys; a numeric segment indexes an array — `items.0.name`); scalars stringify plainly (a JSON `8` → `"8"`, not `"8.0"`), a composite leaf re-encodes to JSON so a `pattern` can still scan it.
- `field` on a `*.headers` is a case-insensitive header name; duplicate headers join with `, ` (RFC 7230).
- optional RE2 `pattern` refines what `field` resolved (capture group 1 else whole match), or scans the raw body when `field` is absent.
- optional `transform` (reuses `ApplyTransform`, which gains its first non-test caller — the "dead mirror" #160 flagged).

Unresolved values are **silently omitted** per SEP-0005, so an incomplete record (bodies not captured) yields fewer properties, never an error. Reserved identity names (`status`/`method`/`duration`) are left to the signal layer and need no `properties:` declaration.

## Scope / follow-ups

- **Lib only** — this is exactly what Subtext pulls; it populates records from its own capture. No CLI/devtools wiring here.
- **Capture side** (populating headers/bodies from the sightmap collector's live CDP) is #130; **devtools wiring** to surface these in `network list|get` follows after.
- The message analogue (exception stacks on `Message` records) is #206.

## Tests

`request_extract_test.go`: body field / nested + array index / numeric stringification, header + pattern, duplicate-header join, pattern-scans-raw-body (form-encoded), transform, composite-leaf-encoded-for-pattern, invalid-pattern-omits, exhaustive silent-omission, `RequestsForRecord` identity+extraction / no-match / no-properties, and the lean-wire guard. `go build/vet/test ./...` green; `gofmt` clean.

### DIFF
--- a/.changeset/request-record-extraction.md
+++ b/.changeset/request-record-extraction.md
@@ -1,0 +1,17 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Observed `Request` records can now carry the payload a `RequestDef`'s
+`properties[]` addresses, and a new matcher resolves them — the runtime half of
+SEP-0005. `Request` gains optional (omitempty) `DurationMs`, `ReqHeaders` /
+`RspHeaders` (`[]Header`), and `ReqBody` / `RspBody` (`*Body`) fields, so a
+producer holding a full capture can populate them while a lazy producer leaves
+them nil and the wire stays lean. `Corpus.RequestsForRecord(rec) []RequestMatch`
+does route+method identity matching (via `RequestsForURL`) and then resolves each
+matched def's `properties[]` against the record — `source` → `field` (JSON
+dot-path with numeric array indexing for a `*.body`; case-insensitive header
+lookup for a `*.headers`) → optional RE2 `pattern` (capture group 1 else the whole
+match, or a scan of the raw source when `field` is absent) → optional
+`transform`. Unresolved values are omitted silently per the spec, so an
+incomplete record degrades to fewer properties rather than erroring.
--- a/go/sightmap/records.go
+++ b/go/sightmap/records.go
@@ -18,9 +18,16 @@ type Message struct {
 	Ts    int64  `json:"ts"` // unix milliseconds
 }
 
-// Request is one observed network request/response. Bodies are not included —
-// they are large and, for a live capture, fetched lazily by the tool that
-// observed the request. Status is 0 until the response is seen.
+// Request is one observed network request/response. Status is 0 until the
+// response is seen.
+//
+// Everything a RequestDef's properties[] can address (the SEP-0005 sources
+// req.body / rsp.body / req.headers / rsp.headers, plus the reserved identity
+// name duration) is carried here, so the record is faithful to what the def can
+// extract. But every such field is optional: a lazy or streaming producer may
+// leave them nil/empty — bodies are large and are often fetched on demand — and
+// property extraction then degrades to "value absent" (silent omission, per the
+// SEP) rather than erroring. A producer holding a full capture populates them.
 type Request struct {
 	Index        int    `json:"index"`
 	Tab          string `json:"tab"`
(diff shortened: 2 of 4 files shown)


nrodd (approved):
LGTM!

PR #208, opened 2026-08-13
sightmap: exception stack matching/extraction on messages (SEP-0006 addendum)

### LINKED ISSUE #206: `messages:` can't match or extract on exception stack traces

joelgwebber:
Closes #206. Stacked on #207 (base `feat/request-record-extraction`) — reuses that PR's `applyPattern` + transform tail. The message analogue of #205's request-body gap, and a small SEP-0006 addendum (no new SEP — it reuses SEP-0005's exact `source`/`field`/`pattern`/`transform` vocabulary).

## Problem

`MessageDef` matched on `level` + a regex over `text`, and the observed record was `Message{Index, Tab, Level, Text, Ts}`. An uncaught exception's stack — its throwing origin and frames — was dropped, so a corpus couldn't classify an exception by where it came from, nor extract the failing location.

## Change

**Record (`records.go`)** — additive, optional:
```go
Message{ …existing… Stack []Frame `json:"stack,omitempty"` }   // empty for a plain console record
Frame{ Function, File string; Line, Column int }
```

**Def + spec (SEP-0006 addendum)** — `MessageDef` gains a stack-addressing `properties[]` mirroring SEP-0005:
- `source: stack` (the only source in v1),
- `field: <frame>.<attribute>` — `<frame>` is `top` (alias for `0`) or a non-negative index, throwing frame first; `<attribute>` is `function`/`file`/`line`/`column`. Required (a structured stack has no meaningful bare-regex scan — the same reasoning that requires `field` for a headers source),
- optional RE2 `pattern` (capture group 1 else whole match) and `transform`.

`schema.md` gets a **Message properties** subsection + the `properties` row + conformance clauses; `sightmap.schema.json` gets a `messageProperty` `$def` (surgical 38-line insertion, style-matched); `docs/reference/schema.md` regenerated.

**Matching (`message_extract.go`)** — `MessageDef.ExtractProperties(rec)` resolves the stack properties, and `Corpus.MessagesForRecord` folds them into `MessageMatch.Properties` — bringing `MessageMatch` to parity with `ComponentMatch`/`RequestMatch` (it previously carried no properties). Unresolved values omit silently, so a plain console record (no stack) extracts nothing.

**Validation (`validate_message.go`)** — `message-property-invalid-name` / `-source-invalid` / `-no-field` / `-pattern-invalid`, mirroring the request-property diagnostics. New conformance fixture `020-message-properties` (ajv + the Go validate harness via a new glob).

## Scope / follow-ups

- **Lib + spec only** — this is what Subtext pulls. Populating `Stack` from the collector's `Runtime.exceptionThrown` `callFrames` (which it already receives and drops) is the capture side, #130.

## Tests

`message_extract_test.go`: top-frame attributes, numeric frame index, pattern (basename capture), exhaustive silent-omission (OOB frame / unknown attr / malformed field / empty function / no pattern match), plain-console-record-no-stack, `MessagesForRecord` folding, and the four validation diagnostics. Full Go suite + `gofmt` clean; `spec` `validate:conformance` + `validate:examples` green.

### DIFF
--- a/go/sightmap/message_extract.go
+++ b/go/sightmap/message_extract.go
@@ -83,9 +83,15 @@ func resolveStackField(frames []Frame, field string) (string, bool) {
 	case "file":
 		return f.File, f.File != ""
 	case "line":
-		return strconv.Itoa(f.Line), true
+		if f.Line == nil {
+			return "", false
+		}
+		return strconv.Itoa(*f.Line), true
 	case "column":
-		return strconv.Itoa(f.Column), true
+		if f.Column == nil {
+			return "", false
+		}
+		return strconv.Itoa(*f.Column), true
 	default:
 		return "", false
 	}
--- a/go/sightmap/message_extract.go
+++ b/go/sightmap/message_extract.go
@@ -1,0 +1,92 @@
+package sightmap
+
+import (
+	"strconv"
+	"strings"
+)
+
+// ExtractProperties resolves d's stack-addressing properties[] against the
+// observed message rec, returning the values that resolved, in declaration
+// order. A property is silently omitted (never an error) when rec has no stack,
+// the addressed frame or attribute doesn't resolve, or a pattern finds no
+// match — the same live-traffic contract as request-property extraction: a plain
+// console record (empty Stack) simply yields nothing.
+//
+// Resolution is source → field → pattern → transform. The only source is
+// "stack"; field addresses a frame and attribute ("top.file", "1.function"),
+// pattern optionally refines the resolved string (capture group 1 else the whole
+// match), and transform post-processes it.
+func (d *MessageDef) ExtractProperties(rec Message) []PropertyValue {
+	if len(d.Properties) == 0 {
+		return nil
+	}
+	var out []PropertyValue
+	for _, p := range d.Properties {
+		if v, ok := resolveMessageProperty(p, rec); ok {
+			out = append(out, PropertyValue{Name: p.Name, Value: v})
+		}
+	}
+	return out
+}
+
+// resolveMessageProperty applies one property definition to rec.
+func resolveMessageProperty(p MessagePropertyDef, rec Message) (string, bool) {
+	var raw string
+	var ok bool
+	switch p.Source {
+	case "stack":
+		raw, ok = resolveStackField(rec.Stack, p.Field)
+	default:
+		return "", false
+	}
+	if !ok {
+		return "", false
+	}
+	if p.Pattern != "" {
+		matched, ok := applyPattern(p.Pattern, raw)
+		if !ok {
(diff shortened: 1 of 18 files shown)


go/sightmap/message_extract.go
```diff
@@ -0,0 +1,98 @@
...
+	if raw == "" {
+		return "", false
+	}
+	return ApplyTransform(raw, p.Transform), true
+}
+
+// resolveStackField resolves a "<frame>.<attribute>" field against a call stack.
+// <frame> is "top" (alias for index 0) or a non-negative integer index;
+// <attribute> is one of function/file/line/column. Returns false when the field
+// is malformed, the frame index is out of range, or the attribute is unknown —
```
nrodd, line 61:
```suggestion
// is malformed, the frame index is out of range, or the attribute is unknown / unset —
```

small nit since known empty attributes also return false

go/sightmap/message_extract.go
```diff
@@ -0,0 +1,98 @@
...
+		matched, ok := applyPattern(p.Pattern, raw)
+		if !ok {
+			return "", false
+		}
+		raw = matched
+	}
+	if raw == "" {
+		return "", false
+	}
+	return ApplyTransform(raw, p.Transform), true
```
nrodd, line 55: Could create a helper for this if you wanted, since some of the logic is shared in request_extract.go:88-98 . Not a big deal though I don't think.

nrodd (approved):
LGTM, just small nits

go/sightmap/message_extract.go
joelgwebber: Thanks, good eye. I'll merge those in a followup.

PR #211, opened 2026-08-13
browser: capture req/rsp headers, duration, and exception stack frames

joelgwebber:
Part of #226 / #170. Stacked on #208 (base `feat/message-stack-extraction`). The **capture side** of the record-faithfulness pair: the reference collector now populates the optional record fields that #207 (Request headers/bodies) and #208 (Message stack) added, so `RequestsForRecord`/`MessagesForRecord` have something to extract from when the reference tooling is the producer.

## Change (`go/browser/collector.go`)

- **Request headers** — `parseRequest`/`parseResponse` now read `request.headers` / `response.headers` off the CDP events (already in hand — no extra round-trip) into `Request.ReqHeaders` / `RspHeaders`. Headers arrive as a CDP object, so they're emitted as a `[]Header` sorted by name (deduped/unordered — the `[]Header` type still supports duplicates for a producer that has the raw header text).
- **Duration** — `applyResponse` sets `DurationMs` to the observed request→response latency.
- **Exception stack** — `parseException` reads `exceptionDetails.stackTrace.callFrames` into `Message.Stack`, throwing frame first; CDP `lineNumber`/`columnNumber` are 0-based and always present, so they're captured as non-nil pointers (a genuine line 0 is preserved, distinct from "not captured").
- `parseResponse` now returns a `responseInfo` struct rather than a 5-value tuple (it was growing a 6th/7th return).

Bodies stay **fetched on demand** (the existing `ResponseBody`/`RequestBody` lazy fetchers) — the handler-side wiring that assembles a full record and renders extracted values is the next slice.

## No output change

The CLI renders selected text fields (`network list`: `method url → status`; `console`: `level text`), not a raw record dump, so `list`/`get` output is unchanged — the new fields ride in the record silently until they're consumed. The follow-up wiring PR surfaces extracted `name=value` pairs.

## Tests

`collector_test.go`: request-header capture, response-header capture (sorted), and exception stack-frame capture including the **captured-zero line/column** case (0-based CDP → preserved as `*0`, not dropped). Full Go suite + `gofmt` green.

Leaves open: eager body retention + surfacing headers in `network get` (#226 — this PR only lands the header/timing *capture*); WebSocket visibility (#130).

### DIFF
--- a/.changeset/collector-capture-headers-stack.md
+++ b/.changeset/collector-capture-headers-stack.md
@@ -1,0 +1,13 @@
+---
+"@sightmap/sightmap": patch
+---
+
+The reference network/console collector now populates the faithful-record fields
+added for SEP-0005/0006 runtime matching. Observed `Request` records carry
+request/response headers (from the CDP `requestWillBeSent`/`responseReceived`
+events — cheap, no extra round-trip) and an observed `DurationMs`; observed
+exception `Message` records carry the `Stack` (from `exceptionThrown`
+`stackTrace.callFrames`, throwing frame first, with 0-based line/column
+preserved). Bodies remain fetched on demand. This is what lets
+`RequestsForRecord` extract body/header properties and `MessagesForRecord`
+resolve stack properties against traffic the reference tooling captured.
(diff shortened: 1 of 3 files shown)


jurassix (approved):
lgtm

PR #212, opened 2026-08-13
cmd/sightmap: surface extracted request/message properties in devtools

### LINKED ISSUE #227: cmd/sightmap: surface extracted request/message properties in devtools
## Summary

Surface extracted request/message property values in the `sightmap` devtools
output. `network list`/`get` and `console list`/`get` should lead each record
with its corpus match (`[Name]` / `[--]`) and trail the extracted `{name=value}`
pairs (plus a `Properties:` block in `get`), by calling `RequestsForRecord` /
`MessagesForRecord` in the devtools handler over the captured record.

This is the CLI-internal consumer of the request-property + message runtime
resolvers whose library halves shipped in v0.23.0. It was deliberately out of
the lib-boundary scope of #170 ("not CLI devtools enrichment"); tracked here for
hygiene.

Implemented by PR #212.

joelgwebber:
Closes the runtime-matching pile: the devtools now **surface extracted property values**, end-to-end. Stacked on #211 (base `feat/collector-capture-headers-stack`).

## Change

**Collector — eager body retention.** Per the "memory isn't a concern here, keep it simple" call, bodies are now coupled to the record rather than fetched lazily at query time:
- request bodies come inline off `requestWillBeSent` (`request.postData`, no round-trip);
- response bodies are fetched on `Network.loadingFinished` — when they're reliably available — in a goroutine **off** the drain loop (a `getResponseBody` round-trip there would stall the cap-8 event channels), gated to XHR/Fetch so page assets (images/fonts/CSS) aren't pulled into memory.

So a buffered `Request` now carries headers (#211) + bodies, and extraction runs against a complete record.

**Handler + render.** `annotateNetwork` switches from `RequestsForURL` (names only) to `RequestsForRecord` (identity + `properties[]` resolved against the live record); `annotateConsole` already gets `MessageMatch.Properties` (stack extraction from #208). Both flow into the entry DTO, and render as:
- a trailing `{name=value, …}` token on `network list` / `console list` lines (classification leads via the `[match]` slot, extracted content trails — mirroring a snapshot's `[Component attr="value"]`);
- a `Properties:` block in `network get` / `console get` detail.

## End-to-end (headless Chrome, local fixture)

A page fetches `/pay.json` (`200 OK`, body `{"status":"declined","amount":42}`) and throws an async exception:

```
network list
[1] [CheckoutPayment]      GET http://localhost:8199/pay.json → 200 OK (Fetch) {outcome=declined, amount=42}

console list
[4] [CheckoutBoom]         exception Error: boom in checkout
    at boomTrigger (<anonymous>:1:118) {origin_fn=boomTrigger}
```

The 200-OK-body-says-declined case is visible; the body was retained from live Chrome and resolved by `RequestsForRecord`. (`origin_file` correctly omitted for the eval'd frame — an anonymous `<anonymous>` URL → empty file → silent omission — while `origin_fn` resolved.) `get` detail shows the `Properties:` block + the retained response body.

## Tests

Unit: `annotateNetwork`/`annotateConsole` extract properties and render (`propsSlot`); existing identity/nil-corpus tests unchanged. Plus the live E2E above. Full Go suite + `gofmt` green; `minor` changeset.

Leaves open under #130: WebSocket visibility.

---

Closes #227 (this PR implements it).
Partially addresses #226 (eager XHR/Fetch response-body retention).
#130 is now scoped to WebSocket visibility only; the HTTP header/body residual was split out to #226.

### DIFF
--- a/.changeset/devtools-property-annotation.md
+++ b/.changeset/devtools-property-annotation.md
@@ -1,0 +1,15 @@
+---
+"@sightmap/sightmap": minor
+---
+
+The `browser start` devtools now surface extracted property values, not just
+matched-def names. The collector eagerly retains request/response bodies for
+XHR/Fetch traffic (request bodies inline from `requestWillBeSent`; response
+bodies fetched on `Network.loadingFinished`, off the event loop), so a buffered
+`Request` record is complete. `sightmap network list|get` annotate via
+`RequestsForRecord` — resolving each matched request's `properties[]` against the
+live headers/body — and `sightmap console list|get` surface a matched message's
+stack `properties[]`. Extracted values render as a trailing `{name=value, …}`
+token on list lines and a `Properties:` block in `get` detail, e.g.
+`GET /api/checkout/pay → 200 OK (Fetch) {outcome=declined}` — the "200 OK but the
+body says declined" case, now visible.
--- a/go/browser/collector.go
+++ b/go/browser/collector.go
@@ -149,6 +149,7 @@ func (c *Collector) attachTab(tabID string) {
 	excCh := conn.Subscribe("Runtime.exceptionThrown")
 	reqCh := conn.Subscribe("Network.requestWillBeSent")
 	respCh := conn.Subscribe("Network.responseReceived")
+	finishedCh := conn.Subscribe("Network.loadingFinished")
 
 	// Enable the domains after subscribing so buffered replays aren't missed.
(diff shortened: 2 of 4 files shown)


nrodd (approved):

jurassix (approved):
lgtm - I stacked pr to flip these network scans to be tail based for consistency with the other call sites. https://github.com/sightmap/sightmap/pull/233

PR #214, opened 2026-08-13
match: chain-matching entry point (MatchChain/NamesForChain/TagsForChain)

joelgwebber:
## What

Adds a chain-matching entry point to `match.Matcher`, for consumers that
classify a **stream of individual observed elements** rather than a full
component tree. Each observed element carries only its own root→leaf ancestor
chain (not a tree), so today there is no first-class API for it — a consumer
reimplements the spec's component identity/tag resolution by hand, and
route-blind.

Three additions on `*match.Matcher`:

- `MatchChain(chain []sightmap.Element, pageURL string) []ChainMatch` — the
  primitive. Builds a single-branch `ComponentNode` spine from the chain, runs
  the existing NFA matcher (`FindAllMatches`) over it, and returns every match
  annotated with the `Depth` of the chain node that completed the selector,
  ordered root→leaf.
- `NamesForChain(...) []string` — the spec's **nearest-enclosing** identity rule
  (schema.md §Tags): the name(s) at the deepest matching level. Almost always a
  single name; returns more than one only on a genuine same-depth identity
  conflict, so the caller sees it rather than a silently-picked winner.
- `TagsForChain(...) []string` — the spec's **tag-union** rule: union of tags
  across every matching level, deduplicated and lexicographically sorted.

Matching is **route-aware** — `pageURL` selects the same view-scoped-plus-global
set `Match` uses, so a view-scoped component applies on the chain exactly as it
would in a full-tree match. This is what makes the downstream consumer's
signal-path naming route-aware in the process.

## Why

Lifts the spec's two component resolution rules into the shared library so they
live once, and unblocks retiring a downstream consumer's parallel single-chain
matcher (FullStory/lidar signal path). Purely **additive packaging** over
already-exported pieces (`ParseQueries` / `Components` / `FindAllMatches`) — no
new matching semantics.

## Notes

- Because the caller supplies only the ancestor chain and not the leaf's own
  subtree, a relational `:has()` on the leaf cannot be satisfied here — the
  inherent limitation of any ancestor-only view. Documented on `MatchChain`.
- Standalone PR off `main` (not part of the #209 stack); `minor` changeset.

## Test

`go test ./match/` — new `chain_test.go` covers nearest-enclosing (incl.
ancestor fallback), tag union (dedup + sort), depth annotation, descendant
selectors attributing to the deepest node, same-depth conflicts, route-aware
view scoping, and the empty/no-match/no-def edges. Full `go test ./...` and
`gofmt -l` clean.

### DIFF
--- a/.changeset/match-chain.md
+++ b/.changeset/match-chain.md
@@ -1,0 +1,16 @@
+---
+"@sightmap/sightmap": minor
+---
+
+`match.Matcher` gains a chain-matching entry point for consumers that classify a
+stream of individual observed elements rather than a full component tree. Each
+observed element carries only its own root→leaf ancestor chain, so there was no
+first-class API for it — a consumer had to reimplement the spec's component
+identity/tag resolution by hand (and route-blind). `MatchChain(chain, pageURL)`
+runs the shared NFA matcher over a single-branch spine built from the chain and
+returns depth-annotated `ChainMatch` values; `NamesForChain` applies the
+nearest-enclosing identity rule and `TagsForChain` the tag-union rule (deduped,
+sorted), so the spec's resolution lives once in the shared library. Matching is
+route-aware, so view-scoped components apply on the chain exactly as in a
+full-tree match. Purely additive packaging over already-exported pieces — no new
+matching semantics.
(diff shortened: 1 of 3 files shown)


sirrah23 (approved):
LGTM!

PR #216, opened 2026-08-14
browser: add --attach degraded mode (attach to a running Chrome) + fix Collector.Stop deadlock

joelgwebber:
## What

Adds `browser start --attach host:port`: the daemon attaches to an
already-running Chrome's CDP endpoint instead of launching (and owning) its own.
This is the basic, opt-in **degraded mode** from the e463/#70 discussion.

- `runAttachedSession` shares the devtools HTTP server + console/network
  collector with the owned-launch path (extracted into a new `startDevtoolsServer`
  helper) but skips the launch/profile/extension/sandbox machinery. The collector
  only ever needed a CDP address, so live console/exception + request matching
  work identically once attached.
- The session is recorded with `Attached: true` and `PID` = the daemon's own pid;
  `browser stop` **detaches** an attached session (SIGTERM the daemon) and
  **never kills** the caller-owned browser.
- Best-effort extension port injection runs off the main path (a browser without
  the sightmap extension makes the service-worker lookup retry for several
  seconds), so it can't delay readiness or block stop.

## What you give up (documented in the flag help + changeset)

- **Pre-attach history** — capture is complete only from attach onward.
  `Network.enable` replays nothing, so traffic/exceptions that fired before the
  daemon attached are gone (`Runtime.enable` does replay buffered console).
- **Profile isolation / clean-tab guarantees** — the browser is caller-owned.
- **Extension** — assumed handled out-of-band; injection is best-effort and may
  no-op if a manually-installed extension has a different ID.

## Drive-by fix: `Collector.Stop` deadlock

Attach surfaced a latent bug: `Stop` cancelled the per-tab drain contexts
*after* `wg.Wait()`. A drain only returns on ctx cancellation, so `Stop` only
ever returned because the owned path kills Chrome first (breaking the CDP
connections out from under the drains). With the browser still alive — exactly
the attach-stop case — the drains never exit and `Stop` deadlocks. `Stop` now
cancels drains *before* waiting and refuses new tab attachments during teardown;
`conn.Close` is nil-guarded. This matters for any library consumer that stops a
collector without tearing down the browser. Regression test added
(`TestCollectorStop_NoDeadlockWhileConnected`).

## Test

- `go test ./...` green; `go vet` + `gofmt -l` clean.
- **E2E** (headless Chrome launched externally on `:9333`, static fixture):
  `--attach localhost:9333` connects and prints its banner immediately; a tab
  opened post-attach yields `console list` → matched `[CheckoutBoom]` exception
  (with stack) and, after a post-attach fetch, `network list` →
  `[CheckoutPayment] GET …/pay.json → 200 OK (Fetch)`. The per-attach limitation
  is visible exactly as predicted: a page-load fetch that beats the 2s tab-poll
  is missed (console survives via replay, network does not). `browser stop` →
  `detached (browser left running)`, daemon exits cleanly (no deadlock), Chrome
  stays up.

Property-*value* annotation (`{outcome=declined}`) still depends on eager body
retention (#212, in review) and is orthogonal to attach.

## Notes

- Standalone PR off `main` (not part of the #209 stack). `minor` changeset.
- Basic scope per the design discussion: assumes a local CDP endpoint; the
  session file's `Port` is the external CDP port (other commands assume
  `localhost:PORT`).

## Testing against an enterprise / SSO app (verified against Salesforce)

The motivating use case: apps behind enterprise SSO / device-trust that
**reject Chrome for Testing**, so you can't log in under the sightmap-managed
CfT. `--attach` solves this — point it at your own branded Chrome, logged in
normally.

macOS gotchas that shape the recipe:
- `--remote-debugging-port` is **launch-time only** (no runtime toggle), and
  Chrome 136+ **silently ignores it on the *default* user-data-dir** — so you
  need a non-default dir.
- CfT can't decrypt branded Chrome's cookies (per-binary macOS Keychain key), so
  copying a profile into CfT logs you out. Use branded Chrome instead.

Working recipe (log in once in a dedicated dir — no copy, no Keychain issues,
doesn't disturb your everyday Chrome):

```sh
# 1. Launch YOUR branded Chrome with CDP on a dedicated, non-default profile dir
"/Applications/Google Chrome.app/Contents/MacOS/Google Chrome" \
  --remote-debugging-port=9999 \
  --user-data-dir="$HOME/.sightmap/profiles/work-debug" &
# 2. Log into the app once in that window (SSO/2FA); it persists across relaunches
# 3. Attach
curl -s http://localhost:9999/json/version   # confirm CDP is actually up
sightmap browser start --attach localhost:9999 --sightmap-dir .sightmap
```

To reuse an existing profile instead of logging in fresh: point `--user-data-dir`
at the **user-data-dir root** (not the profile subfolder — that silently makes a
new empty profile) and select the profile with `--profile-directory="Default"`.

Verified end-to-end against a live `*.lightning.force.com` session: attach
connected, console streamed (buffered messages replayed via `Runtime.enable`),
and post-attach navigation produced Lightning `/aura` XHRs that matched request
defs live; `browser stop` detached and left the Salesforce session untouched.
The pre-attach limitation was visible as designed — the tab loaded before attach
had console (replayed) but no network (`Network.enable` replays nothing).

### DIFF
--- a/.changeset/browser-attach.md
+++ b/.changeset/browser-attach.md
@@ -1,0 +1,19 @@
+---
+"@sightmap/sightmap": minor
+---
+
+`browser start --attach host:port` attaches the daemon (devtools server +
+console/network collector) to an already-running Chrome's CDP endpoint instead
+of launching and owning its own. It is a deliberately degraded mode: no owned
+profile or extension guarantees, capture is complete only from attach onward
+(pre-attach network/console history can't be recovered — `Network.enable`
+replays nothing), and `browser stop` detaches rather than killing the
+caller-owned browser. The collector and devtools query surface are unchanged —
+the collector only ever needed a CDP address — so live console/exception and
+request matching work identically once attached.
+
+Also fixes a latent shutdown-order bug in the collector surfaced by attach mode:
+`Collector.Stop` cancelled the per-tab drain contexts *after* `wg.Wait()`, which
+deadlocked whenever the CDP connections were still healthy at stop time (the norm
+when a consumer stops the collector without tearing down the browser). Stop now
+cancels the drains first and refuses new tab attachments during teardown.
--- a/.changeset/browser-attach.md
+++ b/.changeset/browser-attach.md
@@ -17,3 +17,12 @@ Also fixes a latent shutdown-order bug in the collector surfaced by attach mode:
 deadlocked whenever the CDP connections were still healthy at stop time (the norm
 when a consumer stops the collector without tearing down the browser). Stop now
 cancels the drains first and refuses new tab attachments during teardown.
+
+And fixes extension-server-port injection under attach: `findExtensionSWWS` took
+the *first* extension service worker it found, so in a real browser with several
+extensions installed (the attach case) it injected the port hint into the wrong
+one and the sightmap overlay never learned the server port — falling back to
+probing 7891–7900 and latching onto whatever sightmap server answered first
+(often a different session). It now identifies the sightmap extension by its
+manifest name before injecting, so the correct overlay gets the hint. An owned
+launch never hit this: its isolated profile contains only the sightmap extension.
--- a/go/browser/collector.go
+++ b/go/browser/collector.go
@@ -50,8 +50,9 @@ type Collector struct {
 	consoleDropped int
 	networkDropped int
 
-	tabsMu sync.Mutex
-	tabs   map[string]*tabColl
+	tabsMu   sync.Mutex
+	tabs     map[string]*tabColl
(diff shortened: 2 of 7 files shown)


nrodd:
Random nit but spec/v1/schema.md:562 and the generated docs/reference/schema.md:573 i think can be updated now that things are implemented

nrodd (approved):

jurassix (approved):
lgtm - stacked a simplification that should allow us to drop the `stopping` flag https://github.com/sightmap/sightmap/pull/232

PR #225, opened 2026-08-16
selectors: pierce shadow DOM consistently across live-DOM ops + formalize in spec

### LINKED ISSUE #224: Formalize selector-through-shadow semantics + make live-DOM ops pierce shadow consistently
## Summary

Sightmap selectors pierce shadow DOM when matched **offline** (the captured tree
flattens shadow roots), but every **live-DOM** operation re-found nodes with
`document.querySelector`, which cannot cross shadow boundaries — so a large part
of the tool surface silently disagreed with the corpus on shadow-heavy apps
(Lightning/LWC, and web components generally). The behavior was also never
written down in the spec.

## Background

Standard CSS cannot address shadow content (`>>>`/`/deep/` were removed;
`::part()`/`::slotted()` reach only explicitly-exposed nodes). Sightmap's ability
to name shadow content is therefore a deliberate divergence, achieved implicitly:
`probe.js` walks `shadowRoot` and stamps `data-sightmap-id` on live shadow nodes,
and the extracted tree inlines shadow subtrees as ordinary children. So matching
pierces — but the live lookups that re-find a matched node did not.

## Audit — selector-consuming surface

| Path | Pierced? (before) |
|---|---|
| matcher / snapshot / coverage / capture / `sel-probe` offline | yes (flattened tree) |
| property extraction | **no** — couldn't extract from shadow-DOM leaves |
| interaction (click/fill/hover/value/scroll/wait-for) | **no** — couldn't act on shadow-DOM controls |
| `bounds` / `sel-probe` live count | **no** — false offline/live divergence on shadow |
| `suggest` / `discover` | **no** — missed shadow candidates + shadow nav links |

## Change

1. **Spec** (`spec/v1/schema.md`): formalize the selector/tree model — selectors
   match the flattened captured tree and therefore cross shadow boundaries, a
   deliberate divergence from live `document.querySelector`; a live
   re-implementation MUST traverse shadow roots to agree. Reconcile the
   `exists:`/sub-selector extract modes to the same rule.
2. **Tools**: a shared shadow-piercing resolver (`browser.DeepQueryJS`) backs
   every live lookup (extraction, interaction, bounds, `sel-probe` live count,
   `suggest`/`discover`), so live ops reach shadow content the same way matching
   does.

## Deliberately deferred

- The extension overlay's extraction JS (`content.js`/`resolver.js`) resolves
  its target element directly (already shadow-aware) but its `exists:`/sub-selector
  paths are still light-DOM only. That's a separate in-browser runtime with
  duplicated extraction JS; bringing it to parity pairs with de-duplicating that
  JS and is a follow-up.

joelgwebber:
Selectors pierce shadow DOM when matched offline (the captured tree flattens
shadow roots), but every live-DOM op re-found nodes with `document.querySelector`,
which can't cross shadow boundaries — so property extraction, interaction, `bounds`,
`sel-probe`'s live count, and `suggest`/`discover` were silently shadow-blind and
disagreed with the corpus on shadow-heavy apps. The behavior was also unspecified.

### Fix
- **Shared shadow-piercing resolver** (`go/browser/deepquery.go`, `DeepQueryJS` →
  `__smDeepQuery`/`__smDeepQueryAll`) that walks `document` + all nested
  `shadowRoot`s. Adopted by:
  - property extraction (`observe/properties.go`) — node resolution + `exists:`/sub-selectors
  - interaction (`browser/actions.go`) — `ResolveBySightmapID`, click locator, click-by-selector, value-read, scroll, `wait-for`
  - `bounds` (`cmd_browser_bounds.go`), `sel-probe` live/count paths (`cmd_sel_probe.go`)
  - `suggest`/`discover` scans (`authoring/scan.go`)
- **Spec** (`spec/v1/schema.md`): a new shadow-DOM paragraph in *Selector semantics*
  formalizing that selectors match the flattened tree and cross shadow boundaries
  (a deliberate divergence from live `document.querySelector`), and the
  `exists:`/sub-selector extract rows reconciled to the same rule.

Because capture already stamps `data-sightmap-id` across shadow, the same ids and
selectors now resolve live — no re-capture needed.

### Verification
- Full `go test ./...`, `go vet`, `gofmt` clean.
- Live on a Salesforce Lightning app (shadow-heavy): components inside the
  `one-appnav`/LWC shadow trees that were bare before now extract their props,
  e.g. `[NavItem href="/lightning/o/Account/home" label="Accounts"]`,
  `[FlowInput fieldName="AccountName" type="text"]`.

### Deferred (follow-up)
The extension overlay's `content.js`/`resolver.js` resolve their target element
directly (already shadow-aware) but their `exists:`/sub-selector paths remain
light-DOM only. That separate in-browser runtime has duplicated extraction JS;
bringing it to parity pairs with de-duplicating that JS.

Closes #224

### DIFF
--- a/.changeset/selectors-through-shadow.md
+++ b/.changeset/selectors-through-shadow.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Make selectors behave consistently across shadow DOM. The captured component tree already flattens shadow roots (so offline matching, `sel-probe`'s offline count, and coverage pierce shadow), but every live-DOM operation re-found nodes with `document.querySelector`, which cannot cross shadow boundaries — so property extraction, interaction (click/fill/hover/value/scroll/wait-for), `bounds`, `sel-probe`'s live count, and `suggest`/`discover` were silently shadow-blind and disagreed with the corpus. A shared shadow-piercing resolver (`browser.DeepQueryJS`) now backs all of them, so live operations reach shadow-DOM content the same way matching does (property extraction on shadow-DOM leaves, interaction with shadow-DOM controls, and discovery of shadow-DOM links now work). `spec/v1/schema.md` formalizes the selector/tree model and its shadow-DOM semantics — a deliberate divergence from live `document.querySelector`, which does not cross shadow roots.
--- a/docs/reference/schema.md
+++ b/docs/reference/schema.md
@@ -128,6 +128,7 @@ See [SEP-0002](https://github.com/sightmap/sightmap/blob/main/spec/seps/0002-com
 - A list of strings is tried in order; the first selector that matches wins.
 - Selectors in `children` are evaluated **within** their parent's matched subtree, not globally. This scoping is how Sightmap avoids naming collisions between, say, two different card components that both contain a `button.primary`.
 - Selectors are not required to be unique at their level. If a selector matches multiple elements, all matches are named.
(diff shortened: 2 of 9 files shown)


jurassix (approved):
lgtm - added a stacked pr to extract the deepquery.go to use and embed js file so we can work in native js land; also shifts the algorithm slighly to match `probe.js`. will stack another pr to do this generally.

https://github.com/sightmap/sightmap/pull/234

PR #229, opened 2026-08-16
devtools: serve retained req/rsp bodies + surface headers in network get

### LINKED ISSUE #226: Network collector: reliable HTTP response header/body retention + surface headers in network get
## Summary

The network collector's regular-HTTP request/response header & body handling
needs to be reliable enough to serve as verification evidence and to back
`requests:` property extraction. Split out from #130, which now tracks WebSocket
visibility only.

## Gaps

- **Response bodies aren't reliably retained.** `network get` intermittently
  reports the body was "retained only briefly" / `getResponseBody` fails
  (500 / broken pipe), so a body can't always be inspected or extracted from
  after the fact.
- **Response headers aren't surfaced in `network get`.** They're captured on the
  record but the CLI doesn't display them, making it awkward to confirm a marker
  header.

## Status

Most of the capture already landed: #211 populates req/rsp headers + timing on
the record; #212 eagerly retains XHR/Fetch response bodies. This issue tracks the
residual — reliability (bodies still occasionally unavailable) and header
surfacing — and is needed to properly wrap up the request-property runtime work.

joelgwebber:
Completes the HTTP header/body half of the network-collector work (#226),
stacked on #212.

### Fixes
- **Reliable body reads.** `Collector.ResponseBody`/`RequestBody` now return the
  body eagerly retained on the record at `loadingFinished`, instead of always
  re-fetching via `Network.getResponseBody` — which fails once Chrome has evicted
  the body (`network get` → "response body no longer available" / broken pipe).
  Falls back to a live fetch only when nothing was retained (non-XHR/Fetch types,
  which the eager path skips by design).
- **Surface headers.** `network get` now renders the captured request/response
  headers (already on the record from #211; previously never shown), long values
  truncated.

### Tests
`collector_test.go`: `ResponseBody`/`RequestBody` serve the retained body without
a live fetch (the records carry a zero conn, so a live fetch would fail — a pass
proves the retained copy is used) + a not-found case. `go test`/`vet`/`gofmt`
green.

### Verified
`network get` on a live capture now prints full Request/Response Headers blocks
(confirmed against a running session). The retained-body path is unit-tested.

Closes #226. Stacked on #212 (base `feat/devtools-property-annotation`).

### DIFF
--- a/go/browser/collector.go
+++ b/go/browser/collector.go
@@ -316,9 +316,12 @@ func (c *Collector) Network(f NetworkFilter) ([]sightmap.Request, int) {
 // index, from the collector connection that observed it. Returns (body, found,
 // err); found is false when no such entry is buffered.
 func (c *Collector) ResponseBody(ctx context.Context, index int) ([]byte, bool, error) {
-	reqID, conn, ok := c.lookupRequest(index)
-	if !ok {
+	retained, reqID, conn, found := c.bodyLookup(index, func(r networkRecord) *sightmap.Body { return r.RspBody })
+	if !found {
 		return nil, false, nil
+	}
+	if retained != nil {
+		return retained, true, nil
 	}
 	body, err := getResponseBody(ctx, conn, reqID)
 	return body, true, err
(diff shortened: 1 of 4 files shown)


go/browser/collector.go
```diff
@@ -316,25 +316,52 @@ func (c *Collector) Network(f NetworkFilter) ([]sightmap.Request, int) {
...
+// already retained on it (pick selects ReqBody/RspBody) or the reqID+conn to
+// fetch it live. Preferring the retained copy is what makes `network get`
+// reliable: the eager loadingFinished capture keeps the body in memory, so a
+// query after Chrome has evicted it from its own network cache still succeeds
+// (a live getResponseBody would fail). Falls back to a live fetch only when
+// nothing was retained (e.g. a non-XHR/Fetch type, which wantsBody skips).
+func (c *Collector) bodyLookup(index int, pick func(networkRecord) *sightmap.Body) (retained []byte, reqID string, conn *CDPConn, found bool) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	for i := range c.network {
```
jurassix, line 354: we'll want to update to use `networkIndexByReqID` if we land that PR stack from https://github.com/sightmap/sightmap/pull/233

jurassix (approved):
LGTM - I rebased this one for you bc it had merge conflicts, make sure I didn't foobar anything before you merge

PR #231, opened 2026-08-17
browser: dedupe requestId lookup in the network buffer

jurassix:
Stacked on #212.

applyResponse and retainResponseBody each scanned c.network under c.mu for a matching requestID; retainResponseBody did it twice (once before the getResponseBody round trip, once after). Factored the scan into networkIndexByReqID (tail-first, caller holds c.mu) and call it from all three sites instead of re-scanning inline.

No behavior change: retainResponseBody still re-checks the index after the round trip rather than reusing the pre fetch one, since the buffer can evict or shift entries while the lock is released.


jurassix:
Closing — base got pulled into a native GitHub stack with a diverged branch, blowing up the diff. Reopening as a fresh, unstacked PR with the same commits.

PR #239, opened 2026-08-17
browser: extract embedded eval scripts to real .js files, add Jest coverage

joelgwebber:
> Replaces #236, which GitHub auto-closed when its base branch (`clint/deepquery-shadow-order`, #234) was deleted on merge. Same branch (`clint/embed-browser-js`) and commit — now retargeted directly onto `main`. Continues the stack: this → #237 → #238.

## What this changes

Extracts the browser-side JS embedded as Go string literals in `scan.go`, `actions.go`, `properties.go`, `cmd_browser_bounds.go`, and `cmd_sel_probe.go` into sibling `.js` files, embedded via `go:embed` (the `probe.js` precedent from `go/probe/`). Go call sites now pass dynamic values as JSON-encoded function arguments instead of splicing them into the script text via `fmt.Sprintf`/`strings.Replace`. Adds a Jest suite per extracted file, reusing the jsdom harness added for `deepquery.js` in #234.

## Why

These scripts (some with real loops/conditionals, e.g. `cmd_sel_probe.go`'s element-description walk) lived as unlintable, unformattable, untestable Go string literals — some built via multi-argument `fmt.Sprintf` with placeholders threaded through comments (`// Placeholders: %s = ..., %d = ..., %v = ...`). Moving them to real `.js` files makes them lintable/formattable JS and, paired with the Jest+jsdom harness from #234, actually testable — before this change, no JS in this repo (including `probe.js`) had any test coverage at all.

Two existing tests changed because their fake-CDP handlers distinguished calls by matching a substring against the raw script text (e.g. `"el.value"`, `"elementFromPoint"`) — substrings that used to be unique to one call's self-contained script. Now that each package's shared `.js` file is prepended in full to every call in that file, those substrings appear in every call's script regardless of which function actually runs. Both tests now match on the specific function invocation (e.g. `"\n__smReadValueBySelector("`) instead.

This PR is scoped to the mechanical move only — no JS logic changes, and no reformatting (a separate follow-up branch runs Prettier across everything once, so that diff is reviewable on its own).

### DIFF
--- a/go/authoring/scan.go
+++ b/go/authoring/scan.go
@@ -7,11 +7,15 @@ package authoring
 
 import (
 	"context"
+	_ "embed"
 	"encoding/json"
 	"fmt"
 
 	"github.com/sightmap/sightmap/go/browser"
 )
+
+//go:embed scan.js
+var scanJS string
 
 // Candidate is one selector candidate found by ScanCandidates: a stable
 // data-attribute selector, how many elements it matched, and enough context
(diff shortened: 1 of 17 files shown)


jurassix (approved):
lgtm

PR #242, opened 2026-08-20
fix(extension): fetch corpus directly on http origins so the overlay isn't stuck empty

joelgwebber:
## Problem

The Sightmap Overlay extension renders **0 components** on pages whose corpus the sightmap server serves correctly. `sightmap snapshot` is unaffected.

Root cause, established by instrumenting every layer against a live app:

- In a **single correlated page load**, the extension's service worker **fetches `/sightmap` and receives the full corpus (18 globals + 10 view components)**, yet the content script logs `loaded 0 components` for that same load. The corpus is lost across the `chrome.runtime.sendMessage("fetch-sightmap")` round-trip between `background.js` and `content.js` on a cold / just-woken MV3 worker.
- The failure **sticks for the life of the page**: `version` is a constant per server-instance, so `pollVersion` never refetches, and an open side panel keeps the worker alive so nothing resets. Hard-refresh and extension-reload do not recover it.
- Ruled out: server race (cold tight-poll never saw an empty corpus; the SW itself receives the full corpus), HTTP cache (no cache headers; `no-store` identical), request origin/encoding, side-panel concurrency, foreign-corpus discovery, and content-script non-injection.

## Fix

- The content script now fetches `http://localhost:PORT/sightmap` **directly** on http-origin pages. `host_permissions` already grants `http://localhost:*` and the server sends `Access-Control-Allow-Origin: *`, so the cross-port fetch is allowed. The background service-worker proxy is kept only as an **https mixed-content fallback**.
- `fetchSightmap` treats an **empty corpus as "not loaded"**: it doesn't cache it or record its version, and `pollVersion` refetches while the corpus is empty — so a bad first fetch self-heals instead of sticking.
- Extension manifest bumped `0.1.18 → 0.1.19`.

## Verification

Against a local app, loading the extension from this branch:

- Content script now logs `loaded 28 components` (was `loaded 0`).
- Hovering a card renders the overlay boxes + tooltip with extracted properties (`[MenuCard itemName="…" price="…"]`).
- `prettier --check` clean; `go test ./cmd/sightmap` (Extension/Manifest/Snapshot) passes.
- Plain `sightmap browser start` (embedded extension) re-extracts `0.1.19` and loads the full corpus.

## Follow-ups (not in this PR)

Lower-priority hardening tracked separately: derive the corpus `version` from content/mtime (defense-in-depth), add a corpus-identity check to the extension's `7891–7900` server discovery, and reap leaked `browser start` daemons on abnormal exit.

### DIFF
--- a/.changeset/overlay-direct-fetch.md
+++ b/.changeset/overlay-direct-fetch.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
+Fix the browser overlay extension rendering **0 components** on pages whose corpus the sightmap server serves correctly. The content script proxied its `/sightmap` fetch through the background service worker, and on a cold/just-woken MV3 worker that round-trip intermittently delivered an empty corpus at page load; the empty result then stuck for the life of the page because the per-instance `version` never changes (so `pollVersion` never refetched) and an open side panel keeps the worker alive. `sightmap snapshot` was unaffected (it reads the corpus in-process over CDP). The content script now fetches the local server **directly** on http-origin pages — `host_permissions` already grants `http://localhost:*` and the server sends `Access-Control-Allow-Origin: *` — using the background proxy only as an https mixed-content fallback, and it treats an empty corpus as "not loaded yet" and keeps retrying instead of caching it.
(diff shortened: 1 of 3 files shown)


jurassix (approved):
lgtm

PR #245, opened 2026-08-20
Authoring efficacy: generalized discovery, explain, honest coverage

joelgwebber:
Toolset + skill changes that make from-scratch authoring work on **hook-poor, framework-generated DOMs** (Salesforce Lightning, Angular, LWC/Aura), validated by repeated cold-start authoring runs against a live Lightning app.

## Changes

**1. Generalized selector-candidate generation** (`coverage.SelectorCandidates`)
`gap`/`suggest` previously emitted candidates only from `data-testid`/`data-component`, so on a DOM without test attributes every orphan cluster read `(no stable ancestor) → (no selector hint)` and `suggest` returned "No candidate selectors found." They now rank custom-element tags, design-system classes, ids (`#id` / `[id^=]`), other stable `data-*`, and `name`/`href`/`aria` hooks — data-attributes stay top-ranked but are one input, not an override — dropping only clearly machine-generated tokens (aura ids, digit/hex runs, `word-<hash>` suffixes). `gap` also prints a `container:` hook on the hook-poor path.

**2. New `explain` command** (node-first inspection)
`sel-probe` is selector-first; when you've spotted a node in a captured tree but have no selector yet (common under shadow DOM, where `browser eval`'s `querySelector` returns `[]`), you had to hand-read `*.snap.tree.json`. `explain` picks nodes by selector, `--id`, or `--grep` (role/name), live or offline via `--snap`, and dumps each node's facts, ranked candidates, coverage tier + owning component, and ancestor hooks — shadow-transparent.

**3. Honest coverage**
A broad global backstop component makes every node T1/T2, so a brand-new **unmodeled** page reports "0 orphaned ✓". `snapshot --coverage` and `gap` now warn when a clean pass has zero view-scoped matches ("coverage came entirely from global components").

**4. Authoring skill**
Second mandatory property rule: any selector matching >1 instance must carry a per-instance discriminator (so repeated cards/rows/tabs stay addressable, e.g. `Card[title^="Today"]`). Plus `explain` docs in the tool table + discovery step.

## Testing

`go vet` + full `go test` + Jest (60) green; `gofmt` clean. New unit tests for `SelectorCandidates`, `ViewScopedMatchCount`, the chrome-only advisory, `scan.js` candidates, and `explain` node selection/tier. Behavior verified live on Salesforce Lightning. Changeset: `minor` (new command).

### DIFF
--- a/.changeset/authoring-efficacy.md
+++ b/.changeset/authoring-efficacy.md
@@ -1,0 +1,10 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Authoring efficacy on hook-poor DOMs (Salesforce Lightning, Angular, framework-generated markup):
+
+- **Generalized selector-candidate generation.** `gap` and `suggest` no longer dead-end when a DOM has no `data-testid`/`data-component`. Candidate generation now ranks custom-element tags, design-system classes, ids, other stable `data-*`, and `name`/`href`/`aria` hooks (data-attributes remain the top-ranked input, not an override), dropping only clearly machine-generated tokens. `gap` also emits a container hook on the hook-poor path.
+- **New `explain` command.** Node-first inspection: pick nodes by selector, `--id`, or `--grep` (role/name) — live or offline via `--snap` — and dump each node's facts, ranked selector candidates, coverage tier + owning component, and ancestor hooks. Shadow-transparent (matches the offline matcher), so authors no longer hand-read `*.snap.tree.json`.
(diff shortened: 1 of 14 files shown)


jurassix:
Will get to this one tomorrow morning.

jurassix (approved):
lgtm - stacked some feedback claude found https://github.com/sightmap/sightmap/pull/254

PR #254, opened 2026-08-21
coverage: deterministic candidate ranking + scan.js parity fixes

jurassix:
Two fixes found while reviewing #245: `coverage.SelectorCandidates`'s "other data-*" ranking depended on unsorted Go map iteration order, so tied candidates could reorder (and drop out of the top-4 cap differently) between identical runs; and its `reTrailingNum` regex demoted any id ending in a single digit (`step2`, `tab1`) to a coarser `[id^=]` prefix form, even though `looksHashed`'s own comment says those exact shapes should be spared. Also closes a parity gap in `scan.js`'s `__smBestHook`, which claimed to mirror `SelectorCandidates`' priority but had no fallback for other stable `data-*` attributes or href suffixes.

### DIFF
--- a/go/authoring/scan.js
+++ b/go/authoring/scan.js
@@ -54,10 +54,25 @@ function __smCandidateWorthy(el) {
   return ["a", "button", "input", "select", "textarea"].includes(tag);
 }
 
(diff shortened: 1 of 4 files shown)


joelgwebber (approved):
LG, thanks!

PR #255, opened 2026-08-21
stats: global-row attribution; clearer capture-gate message

### LINKED ISSUE #250: stats: attribute global components in the per-view table
`sightmap stats`'s per-view table shows `Components 0` for a corpus whose components are all declared as flat globals (at the corpus root, not nested under a view). The totals count them, but the per-view rows read 0 and the footer says "N distinct components (per-view rows sum to 0)" — which looks, to an author, like nothing is mapped to the view.

### Ask
Decide and implement the right presentation for global components in the per-view table — e.g. a dedicated `(globals)` row, or attributing each global to the views it can match, or a footer note. Purely a presentation change; the totals are already correct.

### LINKED ISSUE #251: capture: reproduce/repair the first-baseline novelty refusal
Multiple authoring runs reported that `capture` refused to persist the **first** baseline for a view (printing "nothing new vs 1 capture(s) in <view> — not saved", requiring `--force`) when the view's `snapshots/` directory did not yet exist.

The code path (`viewset.Gate`) writes unconditionally when `len(others) == 0`, so on inspection the empty-set case should always persist — which suggests either (a) the reports actually hit a *second* capture in the same session (correct behavior, misremembered), or (b) a real path where a scratch/other capture is counted as an existing one. 

### Ask
Reproduce deterministically. If it's (b), fix so the first real capture always writes and the "nothing new vs N" message never claims a nonexistent prior capture. If it's (a), improve the message so it names *which* capture it compared against (stamp/path) to remove the confusion.

joelgwebber:
Two small authoring-clarity fixes to the offline inventory and the capture gate. Both surfaced when driving the from-scratch authoring loop against a real single-view app whose coverage lives entirely in global components.

## `stats`: attribute globals to their own row (closes #250)

A corpus that maps its coverage with corpus-root **global** components rendered a per-view table that read as empty — every view showed `0` components — with the confusing footer "per-view rows sum to 0". Globals apply to *every* view (`Corpus.ComponentsForRoute`), so folding them into one view's row would misattribute coverage.

Now they get a single leading row and a footer that explains them:

```
 View          Route          Components  Requests
 ──────────────────────────────────────────────────
 (global)      (all views)             3         1
 LightningApp  /lightning/**           0         0
 ──────────────────────────────────────────────────
 1 view  ·  3 distinct components  (3 declared globally, shared by every view)
```

The counts are derived in the CLI from the corpus rather than added to `Stats`, so the `stats --json` contract (and the atlas catalog's `Totals` shape) is unchanged — verified by the existing published-field-set test.

## `capture`: clarify the novelty-gate message (closes #251)

The gate always writes the first capture of a view (`len(others) == 0` short-circuits before the novelty check), so an empty snapshots dir is never refused. But the message — "nothing new vs N capture(s) … not saved" — read like a *first baseline* was being rejected, when in fact a baseline already existed and the new capture was redundant. Reworded to say so plainly:

> capture: `<view>` already has 1 capture(s); this one adds no new component or interactive slot — not saved (use --force to keep it anyway)

Adds a `viewset.Gate` test pinning the first-write guarantee (empty set always writes, even a fingerprint-empty capture; a redundant second is gated with `ComparedTo == 1`), and updates the `sightmap-authoring` skill to match and to reinforce that the first capture always writes.

## Verification
- `gofmt -l` clean; `go build/vet/test ./...` green; `npx jest` green.
- Skill change regenerated into `go/skills/` via `go generate ./skills/...`.

### DIFF
--- a/.changeset/stats-global-row-and-capture-message.md
+++ b/.changeset/stats-global-row-and-capture-message.md
@@ -1,0 +1,9 @@
+---
+"@sightmap/sightmap": patch
+---
+
+Two authoring-clarity fixes for the offline inventory and the capture gate.
+
+`sightmap stats` now attributes corpus-root **global** components and requests to their own `(global) · (all views)` row instead of showing `0` against every view. A corpus whose coverage lives entirely in globals (e.g. a single-view app mapped with global components) previously rendered a per-view table that read as empty, with the confusing footer "per-view rows sum to 0". The globals are shown once, on a leading row, and the footer explains them — globals apply to every view, so folding them into one view's row would misattribute coverage. The `--json` contract is unchanged.
+
+`sightmap capture`'s novelty-gate message no longer reads like it is refusing a first baseline. The first capture of a view always writes (an empty set can't be redundant); reaching the gate means a baseline already exists and the new capture adds nothing, so the message now says so plainly ("<view> already has N capture(s); this one adds no new component or interactive slot — not saved"). The `sightmap-authoring` skill is updated to match and to reinforce the first-capture-always-writes guarantee.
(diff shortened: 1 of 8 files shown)


jurassix (approved):
lgtm - here's a few small changes from my reivew https://github.com/sightmap/sightmap/pull/257

PR #256, opened 2026-08-21
multi-coverage: don't let stale capture dirs manufacture phantom global candidates

### LINKED ISSUE #248: multi-coverage: key cross-view analysis on the [View] header, not the dir name
`multi-coverage` treats each `snapshots/<dir>/` as a distinct view. If two capture directories resolve to the **same** view/route — e.g. a stale directory from an earlier run when the view had a different name, or a stray `*.snap.tree.json` scratch file left in the working directory — the tool manufactures phantom cross-view evidence and recommends promoting that view's own components to globals ("appears in 2+ views → add to components.yaml"). Following that advice wrongly globalizes view-local components.

### Ask
Key the cross-view analysis on each capture's recorded `[View: NAME]` header rather than the directory name, and/or warn when two capture dirs resolve to the same view. Consider ignoring loose `*.snap`/`*.snap.tree.json` files in the working directory (only read `.sightmap/snapshots/`), since `snapshot --out FILE` writing scratch files there is a documented workflow.

joelgwebber:
`multi-coverage` grouped its matrix columns purely by directory name under `snapshots/`. A stale or renamed capture dir — e.g. an old `snapshots/views/` left beside the current `snapshots/home/` for the *same* page — became a second column and manufactured phantom "appears in 2+ views" evidence, so `multi-coverage` advised promoting that page's own view-scoped components to globals (closes #248).

## Fix

A capture dir now counts as a real view only when it matches a view in the **current** corpus (by `SnapBasename` — the name `capture` writes the dir under). Legit captures always match, since `capture` created the dir from that same basename; only stale/renamed/hand-placed dirs fall out.

- **Global-candidate analysis** counts only current columns, so a stale dir can't inflate a component's view count.
- **Matrix** still shows every dir, marking non-current ones with `*` for context.
- **Warning** names the excluded dirs (with their recorded `[View: NAME]` header) and suggests deleting the stale dir or re-capturing under the current name.

Example (old `views/` dir kept beside current `home/`, same page):

```
Component  home  views*
───────────────────────
Banner        1       1
Card          2       2

⚠ 1 capture dir(s) match no view in the current corpus (stale or renamed?) —
  marked "*" above and EXCLUDED from the global-candidate analysis:
  views (recorded [View: OldHome])   → delete the stale dir, or re-capture the view under its current name
```

`Card`/`Banner` are no longer offered as global candidates; before, both were.

## Notes / scope
- The shared `viewset.GroupByView` primitive (used by prune/presence) is unchanged; this is scoped to `multi-coverage`'s promotion analysis.
- A `capture`-side guard against writing a view's snaps under a non-matching dir is a reasonable follow-up but intentionally out of scope here.

## Verification
- New unit tests: current-only candidate counting, stale-column detection, label marker; existing column tests updated for the new field.
- `gofmt -l` clean; `go build/vet/test ./...` green; `npx jest` green; skill regenerated via `go generate ./skills/...`.
- Verified end-to-end against an on-disk fixture (current `home` + stale `views` of the same page): stale column excluded + warned; genuine two-current-view components still promote.

### DIFF
--- a/.changeset/multi-coverage-stale-dirs.md
+++ b/.changeset/multi-coverage-stale-dirs.md
@@ -1,0 +1,7 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`sightmap multi-coverage` no longer manufactures phantom "global candidate" promotions from stale capture directories. It grouped columns purely by directory name under `snapshots/`, so a leftover or renamed dir (e.g. an old `snapshots/views/` kept alongside the current `snapshots/home/` for the same page) became a second column and made that page's own components look like they "appear in 2+ views" — advising the author to wrongly globalize view-scoped components.
+
+A capture dir is now treated as a real view only when it matches a view in the *current* corpus (by `SnapBasename`). Non-current dirs are still shown in the matrix (marked `*`) for context but are excluded from the cross-view global-candidate analysis, and a warning names them so the author can delete the stale dir or re-capture under the current name. The `sightmap-authoring` skill documents the new behavior.
--- a/go/cmd/sightmap/cmd_multi_coverage.go
+++ b/go/cmd/sightmap/cmd_multi_coverage.go
@@ -37,6 +37,16 @@ func runMultiCoverage(args []string) error {
 		globalNames[gc.Name] = true
 	}
 
+	// A capture dir is only a real VIEW if it matches a view in the CURRENT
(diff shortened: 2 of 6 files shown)


jurassix (approved):
LGTM - small NITs stacked https://github.com/sightmap/sightmap/pull/259

PR #262, opened 2026-08-23
browser install: fetch correct arch on linux-arm64 with channel fallback

joelgwebber:
Reported on a linux-arm64 machine: `sightmap browser install` produced

```
~/.sightmap/browsers/chrome-152.0.7977.54/chrome-linux64
```

`chrome-linux64` is the x86-64 Chrome for Testing build, so arm64 hosts were silently getting an x64 Chrome.

## Root cause

`cftPlatform()` in `go/browser/install.go` mapped **every** Linux host to `linux64` regardless of `runtime.GOARCH` — the one platform branch that ignored arch (macOS and Windows already split by arch):

```go
case "linux":
    return "linux64"
```

## The wrinkle

Chrome for Testing does publish a `linux-arm64` platform, but only on the **Beta / Dev / Canary** channels — **not Stable** (verified against `last-known-good-versions-with-downloads.json`). The installer was hardwired to Stable, so simply mapping arm64 → `linux-arm64` would have made install fail outright on arm64.

## Change

- `cftPlatform` now resolves per-arch (`linux-arm64` on arm64, `linux64` on x64); extracted into `cftPlatformFor(goos, goarch)` so the mapping is unit-testable (mirrors the existing `findChrome(goos, …)` pattern).
- `cftRelBin` gains a `linux-arm64` case (`chrome-linux-arm64/chrome`).
- Download resolution now prefers Stable and falls back through **Beta → Dev → Canary** for platforms Stable does not carry (today just linux-arm64). `ResolveLatestStableCfT` is replaced by `ResolveCfT`, which returns the channel it settled on.
- `InstallCfT` prints a one-line warning when it uses a non-Stable channel.

Every mainstream platform still installs from Stable, unchanged. As a side effect, on an arm64 box that previously installed the x64 build, `InstalledCfTPath()` no longer matches the stale `chrome-linux64/` directory, so the next install correctly fetches the arm64 build instead of colliding.

## Tests

New `go/browser/install_test.go` (no network):
- `cftPlatformFor` mapping across OS/arch, incl. the `linux/arm64 → linux-arm64` regression.
- `cftRelBin` for the arm64 path.
- channel resolution: prefers Stable, falls back to Beta for linux-arm64, errors when no channel carries the platform.

`go build ./...` and `go test ./browser/` pass; `gofmt` clean.

## Follow-up

Google has shipped `linux-arm64` to Beta/Dev/Canary for a while; once it reaches **Stable**, arm64 will resolve to Stable automatically with no further change (the warning simply stops firing).

### DIFF
--- a/.changeset/browser-install-linux-arm64.md
+++ b/.changeset/browser-install-linux-arm64.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`browser install` now fetches the correct architecture on **linux-arm64**. It previously mapped every Linux host to the `linux64` (x86-64) Chrome for Testing build regardless of CPU, so arm64 machines silently got an x64 Chrome under `~/.sightmap/browsers/chrome-<ver>/chrome-linux64/`. The platform is now resolved per-arch (`linux-arm64` on arm64, `linux64` on x64). Because Google does not currently ship an arm64 build to the **Stable** channel, resolution falls back through Beta → Dev → Canary for platforms Stable does not carry, printing a one-line warning when it uses a non-Stable channel. Every mainstream platform continues to install from Stable unchanged.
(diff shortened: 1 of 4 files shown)


joelgwebber:
@jurassix Force merging this one, because I need it for testing. Please let me know if you see any issues!

PR #274, opened 2026-08-26
fix(cli): browser bounds accepts the full component-query grammar

### LINKED ISSUE #273: browser bounds takes component names only (not the query grammar), unlike click/fill/hover/wait-for
## `browser bounds` takes component names only, unlike the other interaction commands

`browser bounds` matches its positional argument against component **names** only — exact case-insensitive, or `strings.Contains` with `--substring` (`boundsByComponent` in `go/cmd/sightmap/cmd_browser_bounds.go`). It does **not** parse the argument with `compquery`, so:

- `bounds 'NavLink'` → every `NavLink` instance (name match)
- `bounds 'NavLink[label="Books"]'` → "matched no component" (the whole string is compared as a name)
- `bounds 'ConversationItem MenuButton'` → "matched no component"

Every other interaction command (`click`, `fill`, `hover`, `scroll --component-id`, `wait-for --component`) routes through `compquery.ParseQuery`, i.e. the full grammar with property predicates and descendant chains. The `bounds` usage line calling the argument `QUERY...` (in `sightmap browser --help`) reinforces the wrong expectation — an agent reasonably infers that predicated/descendant queries should work there, and when they don't, over-generalizes to "component queries can't traverse a parent" (they can — via the descendant combinator).

### Suggested fixes

- [ ] **`bounds`: accept the component-query grammar.** Parse with `compquery.ParseQuery` and select with `compquery.FindCandidates`, which returns *all* matches — so `bounds` keeps its multi-match / `--all` semantics while gaining predicates and descendant chains, and becomes consistent with the other commands. (Minimal alternative: rename the arg `QUERY...` → `NAME...` in the usage + flag help and state it's name-only.)
- [ ] **`wait-for` help: document `--component`/`--view`.** The usage line lists only `--url | --selector | --load [--timeout-ms N]`, but `wait-for --component '<query>'` and `wait-for --view <Name>` are implemented and work — they're just missing from `--help`.

joelgwebber:
Closes #273.

`browser bounds` matched component **names** only (`boundsByComponent` did a lowercased exact / `--substring` compare), while every other interaction command (`click`/`fill`/`hover`/`scroll`/`wait-for --component`) routes through `compquery`. So `bounds 'NavLink[label="Books"]'` and `bounds 'ConversationItem MenuButton'` returned "matched no component" — which is a plausible source of the mistaken belief that component queries can't traverse a parent (they can, via the descendant combinator).

## Changes

- **`bounds` default path now uses the component-query grammar.** It parses each positional arg with `compquery.ParseQuery` and selects with `compquery.FindCandidates`, so predicates (`Card[title="X"]`) and descendant chains (`Row Star`) work. `FindCandidates` returns *every* match, so `bounds`'s multi-instance semantics are preserved. Properties are extracted for the queried component names so `[prop]` predicates resolve (mirrors `resolveComponentQuery`).
- **`--substring` unchanged** — still a name-only, case-insensitive fuzzy match (kept as an explicit mode, since the query grammar matches names exactly). **`--all` unchanged.**
- **`wait-for --help`** now lists the existing, working `--component QUERY` and `--view NAME` flags.
- Refreshed the `bounds` usage/flag text (`QUERY` now genuinely means a component query).

Behavior note: the default (non-`--substring`) name match is now case-sensitive/exact via `compquery`, consistent with the other commands (previously it was case-insensitive). `--substring` remains available for fuzzy name lookup.

`go build ./...`, `go vet ./cmd/sightmap/`, `gofmt`, and `go test ./...` all pass. No skill changes (neither skill documents `bounds`); changeset included (`patch`).

### DIFF
--- a/.changeset/bounds-query-grammar.md
+++ b/.changeset/bounds-query-grammar.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`browser bounds` now accepts the full component-query grammar — property predicates and descendant chains — the same engine as `click`/`fill`/`hover`/`wait-for`. Previously it matched component names only, so `bounds 'Card[title="X"]'` or `bounds 'Row Star'` returned "matched no component". Multi-match is preserved (a query returns every matching component's box), `--substring` still does name-only fuzzy matching, and `--all` is unchanged. Also documented the existing `wait-for --component` and `--view` flags in `browser --help`.
--- a/go/cmd/sightmap/cmd_browser.go
+++ b/go/cmd/sightmap/cmd_browser.go
@@ -83,7 +83,7 @@ Interaction (IDs from sightmap snapshot output):
   keypress KEY                        (Enter Tab Escape ArrowUp/Down Backspace ...)
   scroll   [--component-id ID] [--delta-x N] [--delta-y N]
   drag        COMPONENT-ID --delta-x N --delta-y N
-  wait-for    --url PATTERN | --selector SEL | --load  [--timeout-ms N]
+  wait-for    --url PATTERN | --selector SEL | --component QUERY | --view NAME | --load  [--timeout-ms N]
   dialog      accept|dismiss [--text INPUT]
   screenshot  [--out FILE] [--stdout] [--component NAME | --selector SEL] [--expand-pct N]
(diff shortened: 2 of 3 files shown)


jurassix (approved):
lgtm - got a test and some refactors to consider here https://github.com/sightmap/sightmap/pull/275

PR #275, opened 2026-08-26
refactor(cli): dedupe query property extraction, add coverage

jurassix:
resolveComponentQuery and bounds's query-grammar path duplicated the same property-extraction block (queryNames -> relevant -> compByName -> ExtractProperties -> node-id projection); this shares it as extractQueryProperties. Also splits bounds's --all/--substring/query-grammar selection into a pure selectBoundsResults, decoupled from the live browser calls, so it's testable without a CDP connection.

Adds tests for both: TestSelectBoundsResults exercises the Row Star descendant-chain case from #273 directly, plus a [prop] predicate and the --all/--substring modes; TestExtractQueryProperties covers the shared extraction logic.


joelgwebber (approved):

joelgwebber:
Good catch, thanks!

Issue #126, opened 2026-07-31
browser start: agent/headless-friendly launch (detach + auto-handle headless/sandbox + honest status)


joelgwebber:
`browser start` runs in the foreground and launches a daemon that owns Chrome plus the console/network collector for the whole session. Several rough edges make it hard to drive from scripts/agents (one-shot shells) and on headless or sandboxed Linux hosts. Grouping them here as the "make `browser start` script/agent-friendly" problem space.

- [ ] **Foreground blocking → a `--detach` mode.** `start` holds the shell ("Press Ctrl-C to stop"), so an agent whose shell runs each command to completion hangs on it indefinitely. Provide a mode that forks, binds CDP + the HTTP server, prints the tab id, and returns 0.
- [ ] **Daemon reaped between one-shot shells.** Even backgrounded (`nohup … &`), the daemon is reaped when the launching shell tears down, so a fresh-shell-per-command driver loses the session between commands. Fully daemonize (own session/process group via `setsid`/double-fork) so it survives independently of the launching shell.
- [ ] **Headless auto-detection.** With no `$DISPLAY`, Chrome exits "Missing X server or $DISPLAY"; today you must pass `--headless`. Default to headless when neither `$DISPLAY` nor `$WAYLAND_DISPLAY` is set (or at least print an actionable hint).
- [ ] **Sandbox on user-namespace-clamped hosts.** "No usable sandbox" occurs where unprivileged user namespaces are restricted (AppArmor on recent Ubuntu, many container runtimes). `--no-sandbox` is auto-added only for `euid 0` (`finalChromeArgs`), so a **non-root** clamped host fails until the user passes `--chrome-flag=--no-sandbox`. Either detect the clamp and auto-add, or — safer, to avoid silently downgrading the sandbox — make the launch error explicitly suggest the flag. (Sandbox and display are orthogonal; neither is architecture-specific.)
- [ ] **Honest `status`.** `status` can report `running` when the launcher was reaped but an orphaned Chrome survives, because it probes the CDP endpoint rather than the sightmap HTTP server. Make it reflect the true daemon + Chrome + server state.

Launch failures are already surfaced clearly (the launcher tails Chrome's stderr into the error message), so this is about auto-handling and actionable hints, not visibility.

PR #276, opened 2026-08-26
feat(browser): browser start --detach + server-aware status

joelgwebber:
Part of #126 (the `--detach` + honest-status items).

`browser start` is a foreground daemon that owns Chrome + the console/network collector for the session, so it holds the shell — and an agent whose shell runs each command to completion hangs on it forever. `nohup start &` doesn't fix it either: the daemon is reaped when the launching shell tears down.

## Changes

- **`browser start --detach`** re-execs `start` (minus `--detach`) as a background daemon in its **own session** (`setsid`), redirects its output to a log, waits until it is actually serving (session file + CDP + HTTP server), then returns. Because it `setsid`s, it survives the launching shell — the durability `nohup` can't provide. Idempotent if a live daemon already owns the corpus; fails fast with the log tail if the daemon exits early. `--log-file` overrides the default `~/.sightmap/logs/<site>-daemon.log`.
- **Foreground `start`** now says plainly that it is a foreground daemon holding the shell, and points at `--detach` for scripts/agents.
- **`browser status`** now probes the sightmap HTTP server (`:7891`) in addition to Chrome's CDP. When CDP is up but the server has been reaped (Chrome is launched detached and outlives its launcher), it reports `⚠ degraded` with recovery steps instead of a misleading `● running`. `status` now also shows the `server=` port.
- Platform helper `configureDetachedDaemon` (`setsid` on Unix; `DETACHED_PROCESS | CREATE_NEW_PROCESS_GROUP` on Windows).
- Unit test for the arg-stripping helper.

## Verification

Built + vetted + gofmt clean; `GOOS=windows` cross-build passes; `go test ./...` green. Verified end-to-end against headless system Chrome:
- `--detach` returns immediately (shell not held) and the daemon is reachable from a **separate** shell (durability);
- `status` reports `● running server=7891`;
- graceful `SIGTERM` → clean shutdown, no orphan;
- ungraceful `kill -9` of the daemon → `status` correctly reports `⚠ degraded` (CDP up, `:7891` down);
- `stop` reaps the orphaned Chrome, back to `○ no session`.

Follow-ups in the same effort (separate PRs): headless (`no $DISPLAY`) + sandbox (userns clamp) auto-handling, and the skill/docs updates.

### DIFF
--- a/.changeset/browser-start-detach.md
+++ b/.changeset/browser-start-detach.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+`sightmap browser start --detach` runs the daemon in the background (in its own session) and returns once it is serving, so scripts and agents no longer hang on the foreground daemon. Unlike `nohup start &`, the detached daemon survives the launching shell (it `setsid`s into its own session). `browser start` also now prints that it is a foreground daemon holding the shell, and `browser status` probes the sightmap HTTP server in addition to Chrome's CDP — reporting `⚠ degraded` when CDP is up but the server has been reaped, instead of a misleading `running`.
--- a/go/cmd/sightmap/cmd_browser.go
+++ b/go/cmd/sightmap/cmd_browser.go
@@ -70,7 +70,7 @@ Setup:
   install                                                           download Chrome for Testing
 
 Session:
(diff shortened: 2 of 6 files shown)


jurassix (approved):
lgtm

PR #277, opened 2026-08-26
feat(browser): auto-headless on no-display + sandbox hint on launch failure

joelgwebber:
Part of #126 (the headless + sandbox items). Stacked on #276 (base branch `feat/browser-detach-d5ee`).

A fresh `browser start` on a headless Linux host (and on sandbox-restricted distros/containers) fails at Chrome launch until you know the two magic flags. This makes it come up on its own.

## Changes (`go/cmd/sightmap/cmd_browser_start.go`)

- **Headless auto-detection.** On Linux with neither `$DISPLAY` nor `$WAYLAND_DISPLAY` set, `start` now defaults to headless (with a one-line note) instead of dying with "Missing X server or $DISPLAY". An explicit `--headless`, or a real display, skips it; macOS/Windows are unaffected.
- **Sandbox hint.** When Chrome fails to become ready and its stderr shows the zygote-sandbox signature (`No usable sandbox`, etc.) — unprivileged user namespaces clamped by AppArmor on recent Ubuntu or by a container — the launch error now explicitly suggests `--chrome-flag=--no-sandbox`. We deliberately do **not** silently add `--no-sandbox` (it's a security downgrade); we point at it. No hint is shown if `--no-sandbox` is already set or the failure is unrelated.

Sandbox and display are orthogonal, and neither is architecture-specific.

Unit tests added for `sandboxHint` and `shouldAutoHeadless`. `go build ./...`, `GOOS=windows` cross-build, `go vet`, `gofmt`, and `go test ./...` all pass. (The Linux-only runtime paths are covered by unit tests; they can't be exercised live on the macOS dev box.)

### DIFF
--- a/.changeset/browser-start-headless-sandbox.md
+++ b/.changeset/browser-start-headless-sandbox.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`browser start` now comes up out of the box on headless and sandboxed Linux hosts. With no display (`$DISPLAY`/`$WAYLAND_DISPLAY` unset) it defaults to headless instead of dying with "Missing X server or $DISPLAY", and when Chrome's launch fails because the host restricts its sandbox (unprivileged user namespaces clamped by AppArmor or a container) the error points straight at the fix (`--chrome-flag=--no-sandbox`) rather than leaving you to decode Chrome's stderr.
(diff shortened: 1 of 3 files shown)


jurassix (approved):
lgtm

PR #283, opened 2026-08-27
Proposal: Limit component property extraction to component references

joelgwebber:
## What this changes and Why
**Tl;dr:** Component properties in the current spec allow fairly arbitrary extraction by sub-selector/attribute references. This both violates the DRY philosophy (ie, you could define a child component with one selector, and reference it for a parent's property with a different one, because errors or drift); and complicates the extraction code by requiring it to go directly to the underlying DOM for properties, rather than being able to resolve them directly against the Component tree.

This is a _breaking_ SEP, but I believe necessarily-so, and significantly simplifying. I'd strongly prefer to get the model right now, than to try and back out of it later.

I'm also using this opportunity to propose removing the half-baked `transform` operators. We can come back later and rethink these more carefully (eg, perhaps with some more generalizable regex + capture-groups approach).

## Area
- [x] Spec (`spec/`)

## Type of change
- [x] Spec change (requires an accepted SEP — link below)

### DIFF
--- a/spec/seps/0010-tree-closed-component-properties.md
+++ b/spec/seps/0010-tree-closed-component-properties.md
@@ -2,7 +2,7 @@
 sep: 0010
 title: Tree-closed component property extraction
 author: Joel Webber (@joelgwebber)
-status: Draft
+status: Accepted
 created: 2026-08-27
 updated: 2026-08-27
 spec-version-target: 1
--- a/spec/seps/0010-tree-closed-component-properties.md
+++ b/spec/seps/0010-tree-closed-component-properties.md
@@ -1,0 +1,220 @@
+---
+sep: 0010
+title: Tree-closed component property extraction
+author: Joel Webber (@joelgwebber)
+status: Draft
+created: 2026-08-27
+updated: 2026-08-27
+spec-version-target: 1
+related-issues:
+  - https://github.com/sightmap/sightmap/issues/282
+related-discussions: []
+---
+
+## Summary
+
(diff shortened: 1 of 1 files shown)


spec/seps/0010-tree-closed-component-properties.md
```diff
@@ -0,0 +1,220 @@
...
+  selector: 'flowruntime-screen-field'
+  properties:
+    - name: label
+      extract: FlowFormLabel.label
+  children:
+    - name: FlowFormLabel
+      selector: '.slds-form-element__label'
+      properties:
+        - name: label
+          extract: text
```
jurassix, line 67: +1

spec/seps/0010-tree-closed-component-properties.md
```diff
@@ -0,0 +1,220 @@
...
+  - https://github.com/sightmap/sightmap/issues/282
+related-discussions: []
+---
+
+## Summary
+
+Redefine component `properties[]` extraction so it resolves against the abstract
+component tree instead of the live DOM. Every `extract` directive references only
+(a) the matched node's own carried data — its accessible text and its
+attributes — or (b) the extracted property of a component nested beneath it. The
```
jurassix, line 19: this descend-only model works as long as there is common wrapper but there are many weird HTML markups of label-input pairs that are siblings (or distant relatives) and may or may not have a common wrapper compoment.

I think this is fine, and matches how FS does element property defs.

jurassix (approved):
LGTM - ship it

PR #285, opened 2026-08-28
spec: SEP-0010 tree-closed component properties + remove transforms

joelgwebber:
Spec + docs implementation of **SEP-0010 (tree-closed component property extraction)**, now Accepted (#283), whose scope was expanded to also **remove `transform` from every property type**. Spec/docs only — no library behavior change yet (that follows in the implementation herd), so no changeset.

## Tree-closed component properties

- **`spec/v1/schema.md`** — rewrite *Component properties* to the four-mode grammar: `text`, `attr=NAME`, `PATH.prop`, `exists:PATH`. Remove the DOM-shaped modes and the live-DOM "MUST omit offline" clause; state resolution over the complete component tree (descend-only, first-match, silent omission); `attr=`'s observed attribute set is implementation-defined.
- **`spec/v1/sightmap.schema.json`** — narrow `componentProperty.extract`; drop `componentProperty.transform`.
- **`spec/conformance/014-component-properties.fixture`** — rewritten to the new grammar (sub-elements promoted to child components).
- **`spec/seps/0003`** — noted as superseded-in-part (extraction model; declaration shape carries forward).

## Remove `transform` from all property types

Per SEP-0010's expanded scope, the fixed transform vocabulary is removed everywhere, not just from components:

- Drop `transform` from `requestProperty` and `messageProperty` in `sightmap.schema.json`.
- Remove the `transform` rows and the `Transforms` subsection from `schema.md`, plus the message-eval `transform` reference. Request/message extraction keeps `source`/`field`/`pattern`; post-processing folds into `pattern` (RE2 + capture groups).
- `spec/seps/0005` and `spec/seps/0006` get amended-by-SEP-0010 notes.
- `spec/seps/0010` scope broadened (Summary, schema diff, conformance, migration, references).

## Verified

- `go test ./sightmap/ ./match/ -count=1` (incl. spec-conformance + schema tests) passes.
- Docs regenerated (`docs/reference/schema.md`); no `transform` / old-mode references remain in spec or docs.

## Follows separately (herd)

Library (offline extractor + `checkComponentProperties` grammar validation), retiring the live JS extractor **and all `transform` plumbing** (`ApplyTransform`, the req/msg extractor fields + calls, `unknownfields.go` key sets, `stats`), and the authoring-skill rewrite.

### DIFF
--- a/docs/reference/schema.md
+++ b/docs/reference/schema.md
@@ -128,43 +128,32 @@ See [SEP-0002](https://github.com/sightmap/sightmap/blob/main/spec/seps/0002-com
 - A list of strings is tried in order; the first selector that matches wins.
 - Selectors in `children` are evaluated **within** their parent's matched subtree, not globally. This scoping is how Sightmap avoids naming collisions between, say, two different card components that both contain a `button.primary`.
 - Selectors are not required to be unique at their level. If a selector matches multiple elements, all matches are named.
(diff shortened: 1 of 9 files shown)


jurassix (approved):
lgtm

joelgwebber:
Note: I updated this PR to reflect the intended goal of removing `transform:` altogether, not just within component defs.

Issue #282, opened 2026-08-27
Offline matcher: extract component properties: (SEP-0003 evaluation half, component analogue of #170)


joelgwebber:
## Summary

SEP-0003 component `properties:` currently exists in two places but not in the
one that offline consumers depend on:

- **Schema** — accepted into the component schema in #52 (declaration +
  validation).
- **Live CLI** — extracted at runtime by `observe.ExtractProperties` (a batched
  JS pass against a live DOM over CDP), surfaced through `observe.Page(...,
  ExtractProps: true)` and rendered by `render.mergedProps` (`sightmap snapshot`,
  and the browser daemon's snapshots).

What's missing is the **offline evaluation half**: the `match` library never
extracts component properties. `match.Matcher.Match` builds
`ComponentMatch{Name, Memory, Tags}` and never populates `ComponentMatch.Properties`,
and there is no `ComponentDef.ExtractProperties` at all. So a consumer that matches
a **serialized `ComponentNode` tree with no live browser / CDP connection** gets
names, tags, and memory — but no property values — even though the render layer
already knows how to display `ComponentMatch.Properties`.

This is the direct component analogue of #170, which landed the runtime/evaluation
half for `messages:` and request `properties:` but explicitly scoped components
out. #52 gave components the schema; this issue gives them the offline runtime,
symmetric with `RequestDef.ExtractProperties` / `MessageDef.ExtractProperties`.

## Why it matters

The only path that resolves component property values today is the live-CDP one
(`observe.ExtractProperties`). A downstream library consumer that already holds a
matched, serialized component tree — and has no browser to run a live DOM pass —
cannot get component properties at all. The identity dimension (names/tags/memory)
is route-aware and offline via `Matcher.Match`; the property dimension is not,
which is an asymmetry with requests/messages (both of which extract offline from
their observed records).

## Scope

1. **`ComponentDef.ExtractProperties(node *ComponentNode) []PropertyValue`** in
   `go/sightmap`, symmetric with the request/message extractors. Resolve the
   SEP-0003 `extract:` modes over the **serialized node**:
   - `attr=NAME` → `Element.Attrs[NAME]`
   - child CSS selector → walk `Children`
   - `text` / `inner_text` / `text_only` → node text (see fidelity note)
   - `exists:SEL` → presence check over `Children`
   - apply the optional `transform`; silent omission on non-resolution (per the SEP).
2. **Populate `ComponentMatch.Properties`** in `match.Matcher.Match`
[...]

PR #286, opened 2026-08-28
match: resolve component properties offline over the tree (SEP-0010, #282)

joelgwebber:
Library implementation of **SEP-0010** — the offline component-property evaluation half, which resolves issue #282: a consumer holding a matched, serialized component tree (no browser) now gets property values directly. This is the piece FS/lidar pulls back to finish the component-properties arc.

## What changed

- **Offline extractor (`go/match/component_props.go`).** `Matcher.Match` now populates each `ComponentMatch.Properties` by resolving the component's `extract` directives over the matched tree:
  - `text` → the node's accessible name
  - `attr=NAME` → the node's carried attribute
  - `PATH.prop` → the resolved property of a descendant matched component (first-in-document-order per segment)
  - `exists:PATH` → `"true"` if the descendant component matched, else omitted
  Resolution is descend-only (a DAG, so recursion always terminates) and requires **no live DOM**. Unresolved values are dropped silently.
- **Precise def per match.** `MatchQuery` now carries `Def *ComponentDef`, so names/tags/properties come from the exact matched definition. Component names are unique only per parent, so the old `byName[q.Name]` lookups collided — those are removed from both `Match` and `NamesForChain`.
- **`transform` removed from component properties** — dropped from `ComponentPropertyDef`, the raw loader type, `unknownfields`, and `stats`.
- **Validation (`checkComponentProperties`)** — duplicate property names within a component, and unrecognized/removed extract modes (`inner_text`/`text_only`/`inner_html`, bare CSS, mistyped `attr`/`exists`), are now errors.

## Verified

- New tests: `component_props_test.go` (all four forms, silent omission, multi-match first-in-document-order, nested `Price.Amount.text`) and `validate_component_test.go` (valid forms, duplicate names, and each removed/malformed mode).
- `go test ./...`, `gofmt -l`, `go vet` all clean.

## Interim note

The live-DOM extractor (`observe.ExtractProperties` / `properties.js`) still runs on the `snapshot`/browser path and would overwrite these offline values; it does not understand the new `PATH.prop`/`exists:PATH` grammar. Retiring it (and removing the remaining `transform` plumbing from request/message properties) is the next PR in the herd. The offline `Match` path — the one #282 is about — is correct and complete here.

Changeset: `minor` (pre-1.0; new offline resolution + breaking removal of component `transform`).

### DIFF
--- a/.changeset/sep-0010-component-properties.md
+++ b/.changeset/sep-0010-component-properties.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Component properties now resolve offline over the component tree (SEP-0010). `Matcher.Match` populates each `ComponentMatch.Properties` from the matched tree with no live DOM, so a downstream consumer holding a serialized component tree gets property values directly. The `extract` grammar is the four tree-closed forms — `text`, `attr=NAME`, `PATH.prop`, `exists:PATH` — and `transform` is removed from component properties. `sightmap validate` now checks component `properties[]`: duplicate names within a component and unrecognized/removed extract modes are errors.
--- a/go/match/chain.go
+++ b/go/match/chain.go
@@ -61,17 +61,12 @@ func (m *Matcher) MatchChain(chain []sightmap.Element, pageURL string) []ChainMa
 		}
 	}
 
-	byName := make(map[string]*sightmap.ComponentDef, len(entry.components))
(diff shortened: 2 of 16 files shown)


jurassix (approved):
LGTM - couple small changes https://github.com/sightmap/sightmap/pull/288

PR #287, opened 2026-08-28
observe: retire live property extraction + remove request/message transform (SEP-0010)

joelgwebber:
Stacked on #286. Retires the live-DOM property extraction and removes `transform` from request/message properties, per SEP-0010. **Net −571 lines.**

## Kill the live-DOM backflip
Component property values now come *only* from the matcher's offline, tree-closed resolution (landed in #286). The live pass was pure overwrite, so:
- Delete `observe.ExtractProperties` + `observe/properties.js` (+ its JS test).
- `observe.Page` drops the `ExtractProps` option; its matches already carry resolved properties.
- `snapshot`, and the `query`/`bounds` component-query paths, read `ComponentMatch.Properties` directly. `extractQueryProperties` (which built a live-DOM probe set) collapses to a pure projection `queryPropertyValues(matches)`, dropping its `ctx`/`conn`/`matcher`/`pageURL` plumbing.

## Remove request/message `transform`
- Delete `ApplyTransform` (`property.go`) — it had no other callers.
- Drop `Transform` from `RequestPropertyDef`/`MessagePropertyDef`, the raw loader types, the extractor calls, and the `unknownfields` sets. Extraction keeps `source`/`field`/`pattern`; fold post-processing into `pattern`.

## Verified
`go test ./...`, `gofmt -l`, `go vet` all clean.

## One thing I did NOT touch — needs your call
The embedded Chrome **DevTools extension** (`extension/content.js`, `extension/resolver.js`) has its *own* in-page `extractProperties` implementing the now-removed grammar (`inner_text`/`text_only`/`inner_html`/CSS-sub-selector + transforms + `match:`). It is a separate surface — pure in-page JS with no access to the Go matcher — so this isn't a redundant backflip I can just delete; removing it makes the panel stop showing properties, and doing it *right* means porting the tree-closed resolution into the extension. I left it out of this PR (it became stale the moment the SEP-0010 spec merged, so it's pre-existing debt, not introduced here). Options I see: (a) delete the extension's property display now, (b) a dedicated follow-up to port it to the tree-closed model. Flagging for a decision rather than guessing.


jurassix (approved):
lgtm - pushed a small commit to some comments and tests

PR #290, opened 2026-08-28
skill: rewrite authoring property guidance for SEP-0010

joelgwebber:
Rewrites the `sightmap-authoring` skill's component-property guidance for **SEP-0010**. Independent of the library stack (touches only `skills/` + the generated `go/skills/` copy), based on `main`.

## Changes
- **Extract modes** table now lists exactly the four tree-closed forms: `text`, `attr=NAME`, `Child.prop`, `exists:Child`. Removed `inner_text`, `text_only`, `inner_html`, and the raw-CSS-sub-selector row.
- Added the **promote-a-sub-element-to-a-child-component** pattern (with a `ProductCard` → `Title`/`SaleBadge` example) as the way to surface a sub-element value or disambiguate a repeated leaf.
- Removed the **Transforms** subsection entirely (vocabulary + `match:REGEX` + the RE2 guidance) and the `AssemblyCheckbox`/`:has()` transform examples.
- Fixed the "structured data in cards" review-pass example (was `inner_text` + raw CSS) to the child-component pattern.
- Request `properties:` guidance drops the `transform:` bullet and is now `source → field → pattern`.
- Regenerated the embedded `go/skills/sightmap-authoring/SKILL.md` (drift check green).

Changeset: `patch` (user-facing skill content ships in the package).

The `sightmap-browser` skill needed no change — it references "extracted properties" generically and never mentioned transforms or DOM extract modes.

### DIFF
--- a/.changeset/sep-0010-skill.md
+++ b/.changeset/sep-0010-skill.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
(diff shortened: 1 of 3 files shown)


jurassix (approved):
lgtm

PR #291, opened 2026-08-28
extension: port property extraction to SEP-0010 tree-closed model

joelgwebber:
Ports the embedded DevTools extension's component-property extraction to the **SEP-0010** tree-closed model (yak for this was filed off #287). Independent of the library stack and the skill PR — touches only `extension/`, based on `main`.

## What changed
`resolver.js` (the portable core) and its hand-inlined `content.js` copy now resolve the four extract forms over the matched component tree, replacing the removed DOM grammar:
- `text` → the element's DOM text content
- `attr=NAME` → the element's attribute
- `PATH.prop` → a descendant matched component's own extracted property (recursive)
- `exists:PATH` → `"true"` if the descendant component matched, else omitted

Descendant resolution (`firstDescendantEl` → `resolvePath` → `resolveExtract`) finds a child by its component's compound selector scoped to the element via `el.contains`, first in document order — mirroring the Go matcher's "first descendant matched component named X". Removed `inner_text`/`text_only`/`inner_html`/raw-CSS sub-selector and every transform (incl. `match:`). `types.js` PropertyDescriptor typedef updated.

## Deliberate implementation note
`text` resolves to the element's `textContent` — the extension's **implementation-defined accessible text**. A content script has no CDP accessibility tree, so it can't compute the a11y name the Go path uses; SEP-0010 explicitly leaves `text` implementation-defined, so this is conformant (it can differ slightly from the CLI's a11y-name `text`).

## Verified
- No `inner_text`/`text_only`/`inner_html`/`transform`/`match:` references remain in `extension/`.
- `go build ./...` (embed) OK; `node --check` passes for both `content.js` (classic) and `resolver.js` (module).
- Logic mirrors the Go extractor (tested in #286). Live in-panel verification is a manual follow-up (load the extension against a corpus page).

## Sync note
`content.js` remains a hand-inlined copy of `resolver.js` (content scripts run as classic scripts, can't import the module) — the two are kept identical. Unifying them is out of scope here.


jurassix (approved):
lgtm - pushed a test

PR #299, opened 2026-09-02
extract: capture rendered node text as the offline `text` fallback

### LINKED ISSUE #298: match: offline `text` extraction yields empty for role-less nodes
## Problem

The offline property extractor (`go/match/component_props.go`, `resolveExtract`) resolves `extract: text` to `node.Name` — the computed **accessible name**. For a role-less container/leaf (e.g. a custom element or a bare `<span>` with no ARIA role), the accessible name is empty, so a declared `text` property silently resolves to `""` in `snapshot`/`coverage`/`bounds`.

This diverges from how downstream consumers read the same element at runtime (accessible-name → `innerText` → `textContent` fallback), so a `text` property an author pulls up on a role-less node is blank offline but populated live. Authors can't verify these properties from the corpus tooling.

Concretely: a card whose sub-fields live in role-less spans (`.body`, `.code`) can't have those values surfaced by `snapshot`, even though they're the natural discriminators for component queries.

## Cause

`probe.js` already captures `textContent` (truncated) into `ProbeComponent.Text`, but `BuildTree` drops it — `ComponentNode` has no text field — so the offline extractor has nothing to fall back to and returns the (empty) accessible name.

## Fix

1. Carry captured text onto `ComponentNode` (new `Text` field), plumbed in `BuildTree`.
2. `resolveExtract`'s `text` case falls back to the node's text when the accessible name is empty.
3. Normalize component text to a single clean shape (collapse whitespace runs + trim) at the build boundary, so every consumer (snapshot, coverage, bounds, and library consumers) sees the same value.
4. Capture rendered text (`innerText`) rather than raw `textContent` in `probe.js` so offline text matches the rendered/transformed/visible text runtime sees (gated on a perf check — layout is already forced per element during the probe walk).

## Impact

Offline `text` extraction reaches parity with runtime rendered text; role-less card sub-fields become visible and queryable from the corpus tooling. Selectors/coverage are unaffected (they key on tag/id/class/attrs and interactive-node containment, not text).

joelgwebber:
Fixes #298.

## What

Offline `extract: text` resolved to the accessible **name**, which is empty for role-less nodes (custom elements, bare `<span>`s) — so card sub-fields authors pull up (e.g. a flight code, a fare cabin living in plain spans) rendered blank in `snapshot`/`coverage`/`bounds`, even though runtime consumers read the visible text fine. Authors couldn't verify these properties from the corpus tooling.

## Changes

- **`ComponentNode.Text`** — new post-merge field carrying the node's rendered text. Documented in `conformance/component-schema.md`.
- **`BuildTree`** plumbs the probe's captured text onto the node, run through a single `normalizeText` (collapse whitespace runs + trim) at the build boundary, so every consumer sees one clean shape.
- **`resolveExtract`** `text` case falls back to `Text` when the accessible `Name` is empty (Name still wins when present).
- **`probe.js`** captures `innerText` (rendered — reflects `text-transform` + visibility, excludes non-rendered `<style>`/`<script>` and hidden nodes) instead of raw `textContent`; no `textContent` fallback (it re-introduced descendant style/script text), and non-rendered elements are skipped (Chrome's `innerText` on `<head>` quirkily returns its `<style>` text).

## Validation

- Unit tests: Name→Text fallback + precedence (`match`), text capture + normalization + `normalizeText` table (`extract`). Full suite green, `go vet` clean.
- Live on a large SPA (jetblue.com booking results): 105 role-less nodes now carry populated, fully-normalized text, **0** `<style>`/`<script>`/CSS leaks; snapshot probe/extract ~2.7s (no `innerText` reflow blowup — layout is already forced per element during the walk).

## Notes

- Selectors and coverage are unaffected (they key on tag/id/class/attrs and interactive-node containment, not text).
- Captured text is `innerText`-based, so a *container's* `text` aggregates descendant text (bounded by the existing 100-char cap) — extract leaf values via child components (the tree-closed model), as the authoring guidance already recommends.

### DIFF
--- a/.changeset/offline-text-parity.md
+++ b/.changeset/offline-text-parity.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Offline `text` extraction now falls back to a node's rendered text when it has no accessible name, so `extract: text` on role-less elements (custom elements, bare `<span>`s) resolves the visible value instead of silently empty. Component text is captured from `innerText` — rendered, transform- and visibility-aware, excluding `<style>`/`<script>` and non-rendered nodes — and normalized to a single clean shape (whitespace runs collapsed, ends trimmed) at the build boundary, so `snapshot`/`coverage`/`bounds` and library consumers all see the same value the runtime does. Adds a `text` field to the component-node schema.
--- a/go/conformance/component-schema.md
+++ b/go/conformance/component-schema.md
@@ -71,6 +71,7 @@ Viewport bounding box in pixels.
 | Id | string | `id` | pre-merge | Globally unique node ID. Frame-prefixed for sub-frames (`"1_5"`). |
 | Role | string | `role` | post-merge | WAI-ARIA role. `"StaticText"` for text nodes, `"none"` for ignored. |
 | Name | string | `name` | post-merge | Computed accessible name (NOT raw `textContent`). |
+| Text | string | `text` | post-merge | Rendered text content (`innerText`, `textContent` fallback), normalized to a single clean shape (whitespace runs collapsed, ends trimmed). Present for role-less nodes that have no accessible name; the fallback source for `extract: text`. Omitted when empty. |
 | Value | string | `value` | post-merge | Current value for form controls. |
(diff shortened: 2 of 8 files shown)


jurassix (approved):
lgtm

PR #301, opened 2026-09-02
extract: capture the full attribute set for offline matching

### LINKED ISSUE #300: match: offline matcher ignores non-standard attributes (attr selectors + attr= blind)
## Problem

The offline element model only knows the attributes that made it into the probe-generated **selector string** — id, classes, a fixed `keyAttrs` list (`type/name/href/role/for/tabindex/title/alt/placeholder`), all `aria-*`, and all `data-*`. Any other attribute is dropped.

So an attribute like `value` on a non-form element (e.g. react-aria listbox options: `<li role="option" value="JFK">`) is **offline-blind**: `[value="JFK"]` matches nothing offline and `extract: attr=value` renders empty, even though the live DOM (and `sel-probe`'s live query) match fine. `sel-probe` reports the live/offline divergence but authors can't act on it.

## Fix

Capture the full attribute map at probe time (every attribute except `data-sightmap-*`) onto `ProbeComponent`, and merge it into `Element.Attrs` in `BuildTree`. Offline attribute matching (`[attr=...]`, `[attr*=...]`, …) and `extract: attr=NAME` then reach parity with the live DOM for element attributes, and the "attribute not captured" divergence disappears (no warning needed).

## Impact

Offline attr selectors and `attr=` extraction stop silently failing on non-standard attributes. Additive matching capability; existing selectors are unaffected.

joelgwebber:
Fixes #300. **Stacked on #299** (base is `offline-text-parity`).

## What

Offline attribute matching only saw the attributes the probe encoded into the generated **selector string** — id, classes, a fixed `keyAttrs` list, `aria-*`, `data-*`. Any other attribute was dropped, so `[value="JFK"]` and `extract: attr=value` were **offline-blind** on nodes like react-aria listbox options (`<li role="option" value="JFK">`), even though `sel-probe`'s live query matched.

## Changes

- **`probe.js`** captures the full attribute set per element, skipping only injected `data-sightmap-*` ids and framework-internal `_`-scoping attributes (Angular's valueless, per-build-hashed `_nghost-*`/`_ngcontent-*` — pure noise, never a stable selector).
- **`ProbeComponent.Attributes`** carries them; **`BuildTree`/`elementFor`** overlays them onto `Element.Attrs` (raw live values win over the selector-parsed subset).

## Validation

- Unit test: a `value` attr on a `role=option` `<li>` surfaces in `Element.Attrs` though the selector omits it. Full suite green, `go vet` clean.
- Live (jetblue booking results): **406** meaningful non-curated attributes now captured across 56 names (was 617 before filtering out the Angular `_`-scoping noise), **0** leading-underscore. Newly-visible hooks include `jbbutton`, `formcontrolname`, `jbinput`, `autocomplete`, and SVG geometry attrs — useful stable-ish selectors for this app.

## Notes

- Additive matching capability; existing selectors are unaffected.
- The "warn when `attr=` targets an uncaptured attribute" idea is obviated — the offline model now sees them all.

### DIFF
--- a/.changeset/offline-attr-capture.md
+++ b/.changeset/offline-attr-capture.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+The offline element model now captures every attribute the live DOM carries (minus injected sightmap ids and framework-internal `_`-scoping attributes like Angular's `_nghost`/`_ngcontent`), not just the curated subset that landed in the generated selector string. Offline attribute selectors (`[attr=…]`, `[attr*=…]`, …) and `extract: attr=NAME` now match the live DOM for non-standard attributes — e.g. `value` on a `role="option"` `<li>` — closing a source of live/offline divergence.
--- a/go/extract/build.go
+++ b/go/extract/build.go
@@ -155,12 +155,27 @@ func elementFor(pc *ProbeComponent) *sightmap.Element {
 		el = &sightmap.Element{Tag: strings.ToLower(pc.TagName)}
 	}
 
+	// Overlay the full observed attribute set so offline matching sees every
+	// attribute the live DOM carries, not just the subset the generated
+	// selector encoded. Raw live values win over the selector-parsed subset
+	// (they're what a live querySelector would compare against).
+	for name, val := range pc.Attributes {
+		if el.Attrs == nil {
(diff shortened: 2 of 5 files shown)


jurassix (approved):
lgtm

PR #302, opened 2026-09-02
browser: eval awaits a returned promise

### LINKED ISSUE #131: `browser eval` does not await a returned promise
`browser eval` does not await a returned promise: a promise-valued expression comes back as `{}`. Probing anything async (e.g. `fetch(...).then(r => r.status)`) requires firing the work and then reading the result out of band via `network list`, or reshaping the expression to return a synchronous value.

**Expected:** `browser eval` awaits a returned promise and resolves to its value.

joelgwebber:
Fixes #131. **Stacked on #301** (base is `offline-attr-capture`).

## What

`Runtime.evaluate` ran with `awaitPromise: false`, so a `browser eval` expression that evaluates to a Promise — an `async () => {...}()` IIFE, or an awaited `executeTool()` — returned an opaque `{}` instead of the resolved value. Testing async page/tool code required a stash-and-poll workaround (`window.__r = …`; second eval reads it).

## Changes

- **`EvalJSON`**: `awaitPromise: true`. CDP resolves a returned Promise and hands back the settled value (a rejection surfaces as the usual JS exception). No-op for synchronous results, so the many internal callers are unaffected.
- **`runEval`**: bounds the eval with a timeout so a never-settling promise fails cleanly instead of hanging the CLI (`conn.call` already honors `ctx`).

## Validation

- Live: `(async () => { await …; return {awaited:true} })()` now returns `{"awaited": true, …}` (was `{}`); `Promise.resolve(42)` returns `42`. Full suite green, `go vet` clean.

### DIFF
--- a/.changeset/eval-await-promise.md
+++ b/.changeset/eval-await-promise.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`browser eval` now awaits a returned Promise and yields its settled value instead of an opaque `{}`, so async page/tool code (e.g. an awaited `executeTool()`) can be driven directly. A per-eval timeout bounds a never-settling promise. This is a no-op for synchronous expressions.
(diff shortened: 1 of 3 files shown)


jurassix (approved):
lgtm

Issue #135, opened 2026-07-31
Spec/docs reference `fmt` and `check-conventions` commands the CLI does not ship


joelgwebber:
The spec references `fmt --check` / `fmt --write` as normative formatting diagnostics, and `check-conventions` is documented, but both are `unknown command` in the shipped CLI. Official surfaces reference commands that don't exist.

**Expected:** implement the referenced commands, or adjust the spec and docs so they don't advertise unshipped commands.

PR #315, opened 2026-09-03
browser: add mcp list/call for WebMCP tools a page exposes

joelgwebber:
### Summary

Adds `browser mcp list` and `browser mcp call` for the [WebMCP](https://webmachinelearning.github.io/webmcp/) tools a page exposes via `document.modelContext`. A page can declare callable actions to an in-browser agent (`getTools()` / `executeTool()`); these let you enumerate and invoke them from the CLI instead of driving the UI blind.

```
sightmap browser mcp list [--json]
sightmap browser mcp call <tool> [--args JSON] [--json]
```

- **`mcp list`** enumerates each tool's name, description, and input schema (`--json` for full schemas). It distinguishes three states — **native** WebMCP, a **polyfilled** `document.modelContext`, and **absent** — and on absent fails loudly with the Chrome flags that turn native WebMCP on (a UA-provided `getTools` reads as `[native code]`; anything else is treated as a polyfill).
- **`mcp call`** resolves the tool by name and invokes it via `executeTool`, printing the structured result and surfacing any guidance the tool returns. Not-found (lists available names), a thrown tool error, and a missing `document.modelContext` are each reported distinctly rather than as a bare `{}`.

Both run over the existing awaited-`eval` path, so no new transport. One tool runs at a single point in time — WebMCP has no cross-navigation tool response (issue #135) — so if a tool navigates, re-run `mcp list` on the new view.

### Tests

Unit tests cover the pure logic: `--args` validation (empty → `{}`, non-object rejected) and the call-script builder (injects a safe JS string literal for the tool name; embeds the args object). `go build ./...`, `go test ./cmd/... ./browser/...`, gofmt, and the embedded-skills drift check are green.

**Not yet live-verified** against a real WebMCP page in this environment (needs a Chrome session with the native flags, or a page that registers tools). The detection/enumeration/call paths are exercised only via unit tests here; a live smoke test is the remaining check before merge.

### Docs

`browser mcp` reference in `docs/cli/browser.mdx` and rows in the `sightmap-browser` skill (embedded copy regenerated); changeset included (minor).


go/cmd/sightmap/cmd_browser_mcp.go
```diff
@@ -0,0 +1,250 @@
...
+func mcpConnect(addr, tab, sightmapDir string) (*browser.CDPConn, error) {
+	return browser.Connect(resolveCDPAddr(addr, sightmapDir), tab)
+}
+
+// ── list ────────────────────────────────────────────────────────────────────
+
+// mcpListScript enumerates the page's WebMCP tools. It distinguishes absent
+// (no modelContext) from present, and reports whether the implementation looks
+// native (a UA-provided getTools reads as "[native code]") vs a JS polyfill.
+const mcpListScript = `(async () => {
```
jurassix, line 79: optional: we could go:embed this. it's very simple so we can defer till it gets more complex.

go/cmd/sightmap/cmd_browser_mcp.go
```diff
@@ -0,0 +1,250 @@
...
+	fmt.Fprintln(os.Stderr, "(use --json for full input schemas; call one with 'browser mcp call <tool> --args {…}')")
+	return nil
+}
+
+// ── call ────────────────────────────────────────────────────────────────────
+
+// mcpCallScript resolves a tool by name and invokes it via executeTool(tool,
+// args). It reports absent / not-found (with the available names) / a thrown
+// error distinctly, so the CLI can fail loudly rather than returning a bare {}.
+func mcpCallScript(name, argsJSON string) string {
```
jurassix, line 152: same comment here

jurassix (approved):
lgtm

PR #388, opened 2026-09-04
web: add /sightkick, plus a Sightkick docs tab and a real mobile nav

chiplay:
Sightkick had no page on the site. This adds one, documents the CLI, and fixes the nav that the new page pushed on.

## `/sightkick`

The page answers three things in order: what sightkick is, what it is for, and how to start.

- **Hero** — the brand lockup, then the install command and a paste-into-your-agent prompt, both one click. Clicking anywhere in the install field copies it; the trailing glyph swaps to a check.
- **Use cases** — tests that survive a refactor, agentic verification, computer use, agent experience.
- **How it's built** — the four layers, sightmap corpus up through registered WebMCP tools, so the dependency on sightmap is explicit.
- **Get started + CLI reference** — the lower half doubles as the README.

The 3D plates are stills of the `/building` scene captured with a transparent background over CDP, so the model sits on the dark section rather than carrying a cream rectangle with it.

Chapter 07 of `/building` now names sightkick and links here. It no longer reads as `exploratory`, because the tool ships.

## Docs

A `Sightkick` tab with four pages: install and first run, the tool layer's YAML shape, the CLI reference, and running a tool against a live page.

## Nav

Spec and Memory were anchors into homepage sections, and the docs cover both in more depth. They are gone. The prose that introduces Views, Components, Requests and memory now links to those docs pages directly, so the same destinations sit one click from where they are described. The nav is Docs, Sightkick, Atlas, Blog, GitHub.

Below 820px the old rule hid every link but GitHub, which sent a phone visitor to the footer to find anything. There is now a toggle and a full-width panel: Escape closes it, the body stops scrolling behind it, focus lands on the first link.

## Checks

- 159 web tests and 67 root tests pass.
- `pnpm build` clean: 22 pages prerendered, 9 routes declared and all covered.
- Checked at 320, 375, 390, 430 and 1440px. The install command holds one line at every width, and the page has no horizontal scroll.

## Review notes

- `web/public/sightkick/*.webp` are the two captured plates (408K total).
- `SightkickLogo.tsx` is the Figma lockup as outlines with the exported background rects dropped and the fill set to `currentColor`.
- The Developers page is untouched. Its `/api/atlas`, `/openapi.json`, `/api/openapi.yaml` and `/llms.txt` links 404 on the dev server because the `netlify.toml` redirects and the generated `dist/` files only exist after a build. All four return 200 in production.

https://claude.ai/code/session_01KnypdFxyUGtk1jFiw4KKBr


jurassix (approved):
Preview looks good to me!

PR #393, opened 2026-09-04
web: correct chapters 05 and 06 of /building to what sightkick ships

chiplay:
Chapters 05 and 06 of `/building` were written before sightkick had a scenario-testing slice. Both promised things it still doesn't do. This corrects them against `sightmap/sightkick@main` (published CLI 0.5.0) and its `docs/scenario-testing.md`.

## Chapter 05 — self-healing

**Was:** *Tests that find the moved door.* "A test written against the map asks the building where the room went. The selector changes, the name does not, and the run finishes."

Nothing re-resolves a moved element. Re-seeding a corpus is a manual trigger. The run does not finish; it fails.

**Now:** *When a room moves, one line changes.*

> A test pinned to `div.sc-hKgILt` breaks the day that class is regenerated. Point it at `ContinueButton` and the selector lives in one line of the map. Lose the room and sightkick fails the build by name. Edit that line, and every tool above it keeps working.

That claim ships today: the corpus holds the selector once, and `sightkick build` hard-fails on any component reference the corpus has lost. **Badge dropped** — the narrowed claim isn't exploratory.

HUD loses `Runs passed 128 / 128`, which implied an automated healing loop. It now reads `Lines changed 1` and `Call sites touched 0`. Rail label `Self-healing` → `Repairs`.

## Chapter 06 — trajectories

**Was:** "the floors it visits, the rooms it touches, the requests it expects on the way. A trajectory is a route through the building any agent can follow, and any run can be checked against."

Two problems. Sightkick has no request concept at all — its step ops are `navigate | goto | fill | click | waitFor | keypress`, and it never reads a corpus's `requests:`. And "a route any agent can follow" and "any run can be checked against" describe two different artifacts, not one.

**Now:**

> Name the route once: log in, open an item, cart it, check out. Each step becomes a hint on the one before it, so an agent mid-run knows what comes next. Written against a scenario, the route becomes a plan that replays with no agent, and refuses to run once the map moves under it.

A journey compiles to a per-tool "consider this next" breadcrumb (`generator/internal/gen/compile.go:738`). A plan is separate: checked-in JSON, hashed over both the feature text and the compiled IR, replayed by `scripts/run-plan.mjs` with no LLM in the loop.

**Badge kept.** Journeys ship, but plans are example-local — there's no `sightkick plan` or `sightkick run`, and `run-plan.mjs` lives in the repo's `scripts/`, not the npm package.

HUD `Requests expected 3` → `Replay cost: no agent`.

## Also

The markdown twin in `web/scripts/lib/agent.ts` still filed all three "built on top" items under `(exploratory)` after chapter 07 shipped, and the corpus memory note in `web/.sightmap/app.yaml` still said chapters 05–07 carry the badge. Both corrected.

## Length

The first pass ran long. The second commit trims both back into the page's own range so they still scan at a glance:

| Chapter | Before | First draft | Now |
|---|---|---|---|
| 05 | 200 | 354 | 258 |
| 06 | 222 | 372 | 281 |

Neighbours run 213–303. No claim was dropped in the trim.

## Checks

159 tests pass. `pnpm build` clean, 22 pages prerendered.

Chapter `id`s are unchanged, so `#self-healing` and `#trajectories` deep links still resolve.

https://claude.ai/code/session_01KnypdFxyUGtk1jFiw4KKBr

### DIFF
--- a/web/src/components/building/chapters.ts
+++ b/web/src/components/building/chapters.ts
@@ -155,18 +155,17 @@ export const CHAPTERS: Chapter[] = [
   },
   {
     id: 'self-healing',
-    short: 'Self-healing',
+    short: 'Repairs',
     eyebrow: '05 · Built on top',
-    badge: 'exploratory',
-    title: 'Tests that find the moved door.',
+    title: 'When a room moves, one line changes.',
     body:
-      'When a room moves, a test written against selectors walks into a wall. A test written against the map asks the building where the room went. The selector changes, the name does not, and the run finishes.',
+      'Point a test at `div.sc-hKgILt` and it breaks the day that class is regenerated. Point it at `ContinueButton` and the selector lives in the map instead. Move the room, edit that one line, and every tool and test above it keeps working. Sightkick refuses to compile against a room the map no longer has, so a break arrives at build time with a name on it.',
     hud: {
-      title: 'Self-healing run',
+      title: 'The repair',
       rows: [
         ['Room moved', 'ContinueButton'],
-        ['Selectors updated', '1'],
-        ['Runs passed', '128 / 128'],
+        ['Lines changed', '1'],
+        ['Call sites touched', '0'],
       ],
     },
     focus: 'Regression',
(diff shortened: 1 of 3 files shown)


web/src/components/building/chapters.ts
```diff
@@ -179,13 +178,13 @@ export const CHAPTERS: Chapter[] = [
     badge: 'exploratory',
     title: 'Journeys you can name and rerun.',
     body:
-      'Codify a journey on top of the primitives: the floors it visits, the rooms it touches, the requests it expects on the way. A trajectory is a route through the building any agent can follow, and any run can be checked against.',
+      'Name the route once: log in, open an item, cart it, check out. Each step becomes a hint on the one before it, so an agent mid-run knows what comes next. Written against a scenario, the route becomes a plan that replays with no agent, and refuses to run once the map moves under it.',
```
jurassix, line 181: Image

the last line doesn't seem like its coherent? 

> becomes a plan that replays with no agent, and refuses to run once the map moves under it.

maybe we it should say:
> becomes a plan that replays with no agent, and continues to run once the map moves under it.

jurassix (approved):

PR #403, opened 2026-09-07
skill(authoring): reframe property principles around purpose (kill text/label noise)

joelgwebber:
Rewrites the sightmap-authoring skill's **Property extraction principles** so agents stop generating noise properties.

## Why
The old Rule 1 — *every link/button must have at least one property* — was satisfied by `label: text` busywork (`FeedbackButton label="Give Feedback"`), because it demanded *a* property, not a *useful* one. `text` defaults to the accessible name the snapshot already prints, so a lone restating property is provably redundant.

## What
- **Lead-in:** a property earns its place two ways — **discriminate** an otherwise-ambiguous instance so a query can address it, or carry **signal** a downstream event/agent needs. Anything else is noise: a `text`/`label` restating the accessible name, or a `text` on a nameless container (whole-innerText subtree dump).
- **Rule 1** now requires a link/button to be **identifiable** (by its accessible name — a self-naming button needs no property — or a useful property), never a name restatement.
- **Rule 2** (per-instance discriminator for repeated components) preserved, with a carve-out: *responsive duplicates* (same value on every match, one visible) → narrow the selector, don't invent a discriminator.

## Validation
Applied to the JetBlue corpus (4 views) via an agent following this wording: it drove ~90% of property decisions cleanly and removed **21 noise properties with zero tool breakage**; coverage held at 0 orphaned. The edge cases it surfaced are captured as design refinements (not folded here, to keep the rules terse).

Embedded `go/skills/` copy regenerated (no drift); `go build` green; `patch` changeset included.

### DIFF
--- a/skills/sightmap-authoring/SKILL.md
+++ b/skills/sightmap-authoring/SKILL.md
@@ -354,11 +354,14 @@ components:
 
 A property earns its place two ways: it **discriminates** an otherwise ambiguous
 instance so a query can address one (`Card[title^="Today"]`), or it carries
-**signal** a downstream event or agent needs (`price`, `status`, `sku`). A
-property that does neither is noise: a `text`/`label` that just restates the
-node's accessible name (the snapshot prints that anyway), or a `text` on a
-container with no accessible name (it resolves to the whole innerText — a subtree
-dump, not a value; promote the real value to a child instead).
+**signal** a downstream event or agent needs (`price`, `status`, `sku`) —
+including a control's current **state** (a sort's active option, a passenger
+count, a field's value), which counts even when it equals the accessible name;
+the test is whether the value would read differently in another capture, not
+whether it matches the name. A property that does neither is noise: a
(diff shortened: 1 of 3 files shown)


skills/sightmap-authoring/SKILL.md
```diff
@@ -384,6 +402,15 @@ Two property rules are **mandatory**:
       extract: attr=aria-label
 ```
 
+A `label` restating the accessible name is noise **only when the component's
+selector specifically identifies that control**. If the selector is generic (a
+utility class matching a category), the component is effectively generic and the
+label is its discriminator — keep it, but name the component for what the
+selector actually matches (`Button[label=…]`), not a specific thing the selector
+can't back up (`ExploreMoreFlights` on `a.rounded-button`). Fixing the selector
+to a stable, specific hook is the better end state; until then, an honest generic
```
jurassix, line 411: that's nice

jurassix (approved):

PR #411, opened 2026-09-08
fix: live/offline selector-matcher parity (:not(:has()) + form named-property collision)

### LINKED ISSUE #410: Selector matching diverges from the browser — offline and live — on real, form-heavy pages
## Summary

Neither of Sightmap's two selector matchers agrees with a real browser on pages that use standard CSS relational selectors and/or contain forms. The two matchers diverge from the browser in **opposite directions**, so on a given corpus **neither is a trustworthy oracle** — which corrupts both the authoring discipline (`sel-probe` before writing YAML) and any coverage/validate signal built on the offline matcher. Surfaced by dogfooding the JetBlue corpus.

## 1. Offline matcher: `:not(:has())` and a combinator inside `:not()`

The offline component-tree matcher (used by `snapshot`, `coverage`, `validate`, `sel-check`, and the overlay) mishandles two standard CSS Selectors Level 4 constructs:

- **`:not(:has(...))` / `:is(...:has(...))`** — the relational pseudo nested inside `:not()`/`:is()` was evaluated by the *flat* element matcher, which ignores `:has()`. So the leaf-scoping pattern `X:not(:has(X))` matched **zero** nodes offline while resolving fine live. Concretely, `jb-form-field-container:not(:has(jb-form-field-container))` (the central checkout form-field component) was **18 live / 0 offline** — coverage/validate were blind to it.
- **A descendant/child combinator inside `:not()`** — e.g. `button:not(nav button)` — failed to **parse**, so a corpus using valid CSS did not `validate` even though it resolved live.

This blocks normalizing on the `X:not(:has(X))` leaf-scoping pattern, which we want to lean on (standard CSS relational pseudos) instead of a bespoke structural-identity DSL.

## 2. Live deep query: under-reports on pages with a form

The live shadow-piercing query `__smDeepQueryAll` (behind `sel-probe`, component-query `click`/`fill`/`hover`, and live property extraction) walked the DOM via instance properties (`node.children`). On an `HTMLFormElement`, a control whose `name`/`id` is `children` makes the form's named getter return that control instead of the element `HTMLCollection` (`[LegacyOverrideBuiltIns]`), so `for (const c of node.children)` throws "children is not iterable" on that form. The blanket `catch` then returns a **truncated** result — silently `0` matches for any element after the form in document order.

On jetblue.com's home page (the booker form has a `children` passenger-count control), `sel-probe` reported **0** for present, visible elements — `[data-fs-element="submit"]`, `a.rounded-button`, `[data-fs-element="points-toggle"]`, `[data-fs-element="dy-container"] a`, and more — while both `querySelectorAll` and the offline matcher saw them.

## Why it matters together

- Checkout: **live `sel-probe` = 18 / offline = 0** (via `:has`, issue 1).
- Home: **`sel-probe` = 0 / offline = N** (form collision, issue 2).

The divergences run both ways, so during the JB passes only `browser eval` + native `querySelectorAll` was a reliable count oracle. Closing both restores parity so authors and tooling can trust the matchers again.

## Fix

Both issues are pure parity bugs against standard CSS / DOM (no spec change):

1. Generalize the offline `:not()` argument to a complex-selector list and make node matching ancestor-chain aware, so `:has()` nested inside `:not()`/`:is()` and a combinator inside `:not()` are honored. Sibling combinators stay rejected loudly at parse. Adds conformance fixtures `013-not-has-leaf-scoping` and `014-not-complex-descendant`.
2. Traverse the live DOM via cached `Node.prototype`/`Element.prototype` accessors, which bypass named-property shadowing.

joelgwebber:
Closes #410.

Brings both selector matchers back into parity with the browser on real, form-heavy pages. Two independent parity bugs, one per commit; both are pure standard-CSS / DOM fixes (no spec change).

## 1. `fix(match)` — offline `:not(:has())` + combinator inside `:not()`

- `:not(:has(...))` / `:is(...:has(...))` evaluated their argument via the flat element matcher (ignores `:has()`), so `X:not(:has(X))` leaf-scoping matched **0** offline (`jb-form-field-container:not(:has(jb-form-field-container))` was 18 live / 0 offline).
- A descendant/child combinator inside `:not()` (`button:not(nav button)`) failed to **parse**, so valid CSS wouldn't `validate`.

Generalizes `SelectorPart.Not` to a complex-selector list and makes node matching ancestor-chain aware (`MatchesNodeChain`; the NFA threads the root→node chain) so a combinator-`:not` evaluates its ancestor constraint and `:is`/`:not` arguments are evaluated node-aware. Sibling combinators stay rejected loudly at parse.

Tests: `match/not_test.go`, `sel_parse` tests, and conformance fixtures `013-not-has-leaf-scoping` / `014-not-complex-descendant` (the cross-implementation contract).

## 2. `fix(browser)` — live deep query under-reports on forms

`__smDeepQueryAll` walked via `node.children`. On a `<form>` with a control named `children`, the form's named getter returns that control instead of the `HTMLCollection` (`[LegacyOverrideBuiltIns]`), so `for (const c of node.children)` throws and the blanket `catch` returns a **truncated** result — silently 0 for any match after the form. Now traverses via cached `Node.prototype`/`Element.prototype` accessors (cached inside the function, since the file is re-eval'd in one tab context). Fixes `sel-probe` and component-query `click`/`fill`/`hover`.

Test: `browser/deepquery.test.js` (named-property-collision regression).

## Verification

- Full Go suite + `deepquery` jest green; gofmt clean; embedded-skills drift check clean.
- `sightmap validate` now passes on the JetBlue corpus (previously errored on the combinator-`:not`).
- Live on jetblue.com: `sel-probe` now equals native `querySelectorAll` for every previously-flagged selector (submit 1/1, input-search a 1/1, rounded-button 1/1, points-toggle 2/2, dy-container a 3/3, dy_cta 1/1).

Changesets: `minor` (offline matcher; `SelectorPart.Not` type change) + `patch` (live deep query).

### DIFF
--- a/go/browser/deepquery.js
+++ b/go/browser/deepquery.js
@@ -4,19 +4,51 @@
 // own light-DOM subtree, fully, before that node's shadow-DOM subtree). See
 // deepquery.go for why this exists and schema.md's "Selector model & shadow
 // DOM" for the flattening rule this mirrors.
-
+//
+// DOM access goes through cached prototype accessors, NOT instance properties
+// (node.children / el.matches / el.shadowRoot). An <input name="children"> makes
+// HTMLFormElement's named getter return that control for form.children instead
+// of the HTMLCollection, so a naive `for (const c of node.children)` throws
+// "children is not iterable" on that form and the blanket catch below then
+// returns a TRUNCATED result — silently 0 for any match after the form. (JetBlue's
+// booker form has a "children" passenger-count control; that one collision made
+// sel-probe and click/fill under-report.) The cached prototype accessors bypass
+// any such named-property shadowing entirely.
+//
+// The accessors are cached INSIDE the function (not at top level): this file is
+// prepended to CDP evals that run repeatedly in one tab context, so it must
+// declare no top-level lexical bindings (a second eval of a top-level `const`
+// throws "already declared"). Function declarations are re-eval-safe; keep it
+// that way. See deepquery.go.
 function __smDeepQueryAll(root, sel) {
+  const childNodesOf = Object.getOwnPropertyDescriptor(
+    Node.prototype,
+    "childNodes",
+  ).get;
+  const matchesFn = Element.prototype.matches;
+  const shadowRootGetter = Object.getOwnPropertyDescriptor(
+    Element.prototype,
+    "shadowRoot",
+  ).get;
+  // shadowRoot's getter lives on Element.prototype, so it is only valid for
+  // element nodes (a Document/ShadowRoot root has none).
+  const shadowOf = (node) =>
+    node.nodeType === 1 ? shadowRootGetter.call(node) : null;
+
   const out = [];
   function visit(node) {
-    for (const child of node.children) {
-      if (child.matches(sel)) out.push(child);
+    for (const child of childNodesOf.call(node)) {
+      if (child.nodeType !== 1) continue; // ELEMENT_NODE only
+      if (matchesFn.call(child, sel)) out.push(child);
       visit(child);
-      if (child.shadowRoot) visit(child.shadowRoot);
+      const sr = shadowOf(child);
+      if (sr) visit(sr);
     }
   }
   try {
     visit(root);
-    if (root.shadowRoot) visit(root.shadowRoot);
+    const rootShadow = shadowOf(root);
+    if (rootShadow) visit(rootShadow);
   } catch (e) {
     return out;
   }
--- a/.changeset/quiet-forms-deepquery.md
+++ b/.changeset/quiet-forms-deepquery.md
@@ -1,0 +1,9 @@
+---
+"@sightmap/sightmap": patch
+---
+
+Live shadow-piercing query (`sel-probe`, `click`/`fill`, property extraction) no longer under-reports on pages with a form.
+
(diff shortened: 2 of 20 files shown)


go/browser/deepquery.js
```diff
@@ -4,19 +4,51 @@
...
+// booker form has a "children" passenger-count control; that one collision made
+// sel-probe and click/fill under-report.) The cached prototype accessors bypass
+// any such named-property shadowing entirely.
+//
+// The accessors are cached INSIDE the function (not at top level): this file is
+// prepended to CDP evals that run repeatedly in one tab context, so it must
+// declare no top-level lexical bindings (a second eval of a top-level `const`
+// throws "already declared"). Function declarations are re-eval-safe; keep it
+// that way. See deepquery.go.
 function __smDeepQueryAll(root, sel) {
```
jurassix, line 23: we could also use treewalker here if this gives us more trouble in the future

jurassix (approved):
lgtm

go/browser/deepquery.js
joelgwebber: Ooh, good call. I'll land this, but do a followup to see if we can simplify all this noise.

PR #430, opened 2026-09-09
fix(browser): name the real Chrome flag for native WebMCP

### LINKED ISSUE #413: WebMCP error suggests Chrome flags that don't enable native WebMCP
Spent yesterday putting sightmap and sightkick through a real evaluation — mapped one view of a small Astro app on Netlify, wrote three read-only tools, ran them both ways. Hit this trying to get `--via webmcp` going:

```
✗ sightmap browser: no WebMCP tools here: document.modelContext is not available on this page.
  Native WebMCP needs Chrome flags — start the session with:
    sightmap browser start \
      --chrome-flag=--enable-blink-features=ModelContext,ModelContextTesting \
      --chrome-flag=--enable-features=DevToolsWebMCPSupport
  A page can also register tools via its own script; if you expect tools here, confirm it ran.
```

So I restarted the session with both of those. Still no `document.modelContext`. Tools only showed up after injecting the runtime, and `sightmap browser mcp list` reported them as `WebMCP (polyfilled)`.

From that I concluded Chrome didn't have a native WebMCP surface at all yet, and wrote it into my evaluation as a finding — that the WebMCP path is entirely sightkick's polyfill, and the whole browser story is a bet on an API that hasn't shipped.

That conclusion was wrong. The error message is what led me there.

## The flag that actually works

`implementation-status.md` in webmachinelearning/webmcp points at `chrome://flags/#enable-webmcp-testing`. The command-line form is `--enable-features=WebMCPTesting`. With that instead:

```
$ sightmap browser start --detach --chrome-flag=--enable-features=WebMCPTesting --url 'https://example.com'
$ sightmap browser eval '"document.modelContext=" + typeof document.modelContext + " | ua=" + navigator.userAgent.match(/Chrome\/[0-9.]+/)[0]'
"document.modelContext=object | ua=Chrome/152.0.0.0"

$ sightmap browser eval 'document.modelContext.constructor.name + " :: " + Object.getOwnPropertyNames(Object.getPrototypeOf(document.modelContext))'
"ModelContext :: ontoolchange,executeTool,getTools,registerTool,constructor"
```

That's example.com. No sightkick runtime on it, nothing injected. A real native `ModelContext` with the spec's methods on its prototype.

Then loading my own app, which ships the runtime as two script tags:

```
$ sightmap browser eval '"polyfill=" + ("__sightkickPolyfill" in document.modelContext)'
"polyfill=false"

$ sightmap browser mcp list
WebMCP (native) — 3 tool(s):
  count_saved_pages …
  list_coloring_pages …
  get_page_visibility …

$ sightmap browser mcp call count_saved_pages --args '{}'
{ "ok": true, "value": "31 pages saved", "guidance": [ … ] }
```

Native surface, tools registered on it, calls returning. Works end to end.

## The fix

Swap the flags in that error string for `--enable-features=WebMCPTesting`, and maybe mention `chrome://flags/#enable-webmcp-testing` as the UI equivalent. One line.

Worth doing soon, I think — anyone evaluating this right now follows those flags, watches them fail, and draws the same wrong conclusion I did. Which sells the project short, because the native path works today.

## Two things that went the other way

Since I'm filing a complaint, the balance: the runtime registers on `document.modelContext`, which is the current spec surface after it moved off `navigator`. A lot of WebMCP tooling still hasn't followed. And when native support is present the runtime defers to it instead of clobbering it — `polyfill=false` above. That's a real progressive-enhancement shim, not a permanent dependency, and it changed my read of the project quite a bit.

**Environment:** `@sightmap/sightmap` 0.31.2, `@sightmap/sightkick` 0.6.0, Google Chrome 152.0.7977.76, macOS 26.6.2.

chiplay:
The "no WebMCP tools here" error told users to start Chrome with `--enable-blink-features=ModelContext,ModelContextTesting` and `--enable-features=DevToolsWebMCPSupport`. None of those are Chromium features. Chrome ignores unknown names without a warning, so on Google Chrome `document.modelContext` stayed undefined after following the message, and the native path looked unshipped.

The native surface is behind the `WebMCP` runtime feature. `--enable-features=WebMCPTesting` turns it on and is the command-line form of `chrome://flags/#enable-webmcp-testing`. The error text and the `browser mcp` docs now name that flag.

The wrong flags went unnoticed because the managed Chrome for Testing has WebMCP on by default through Chromium's field-trial testing config, so they appeared to work there.

Verified on Google Chrome 152.0.7977.83 with a fresh profile on example.com:

| Flags | `document.modelContext` |
|---|---|
| none | undefined |
| old flags from the error | undefined |
| `--enable-features=WebMCPTesting` | native `ModelContext` |

Fixes #413

https://claude.ai/code/session_014Nj6NtC6kV551kYqJXgaev

### DIFF
--- a/go/cmd/sightmap/cmd_browser_mcp.go
+++ b/go/cmd/sightmap/cmd_browser_mcp.go
@@ -55,15 +55,16 @@ type mcpCallResult struct {
 }
 
 // mcpAbsentError is the loud, actionable message for a page with no WebMCP
-// surface — the "absent" arm of the native/polyfilled/absent detection. Native
-// WebMCP is behind Chrome flags, so it names them; a page can also register
-// tools via its own script.
+// surface — the "absent" arm of the native/polyfilled/absent detection. Google
+// Chrome ships native WebMCP off, behind chrome://flags/#enable-webmcp-testing
+// (the WebMCPTesting feature), so the message names that flag; Chrome for
+// Testing has it on by default via the field-trial testing config. A page can
+// also register tools via its own script.
 func mcpAbsentError() error {
 	return fmt.Errorf("no WebMCP tools here: document.modelContext is not available on this page.\n" +
-		"  Native WebMCP needs Chrome flags — start the session with:\n" +
-		"    sightmap browser start \\\n" +
-		"      --chrome-flag=--enable-blink-features=ModelContext,ModelContextTesting \\\n" +
-		"      --chrome-flag=--enable-features=DevToolsWebMCPSupport\n" +
+		"  Google Chrome ships native WebMCP off — start the session with:\n" +
+		"    sightmap browser start --chrome-flag=--enable-features=WebMCPTesting\n" +
+		"  (the same switch as chrome://flags/#enable-webmcp-testing).\n" +
 		"  A page can also register tools via its own script; if you expect tools here, confirm it ran.")
 }
 
--- a/.changeset/webmcp-flag-hint.md
+++ b/.changeset/webmcp-flag-hint.md
@@ -1,0 +1,11 @@
+---
+"@sightmap/sightmap": patch
+---
+
+`browser mcp` no longer points at Chrome flags that do not exist. The "no WebMCP
+tools here" error used to suggest `--enable-blink-features=ModelContext,ModelContextTesting`
+and `--enable-features=DevToolsWebMCPSupport`; Chrome silently ignores those
+names, so following them left `document.modelContext` undefined on Google Chrome
+and made the native path look unshipped. The message and the `browser mcp` docs
+now name the real switch, `--enable-features=WebMCPTesting` (the same feature as
+`chrome://flags/#enable-webmcp-testing`). Fixes #413.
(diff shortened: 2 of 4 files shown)


go/cmd/sightmap/cmd_browser_mcp.go
```diff
@@ -55,15 +55,16 @@ type mcpCallResult struct {
...
+// Testing has it on by default via the field-trial testing config. A page can
+// also register tools via its own script.
 func mcpAbsentError() error {
 	return fmt.Errorf("no WebMCP tools here: document.modelContext is not available on this page.\n" +
-		"  Native WebMCP needs Chrome flags — start the session with:\n" +
-		"    sightmap browser start \\\n" +
-		"      --chrome-flag=--enable-blink-features=ModelContext,ModelContextTesting \\\n" +
-		"      --chrome-flag=--enable-features=DevToolsWebMCPSupport\n" +
+		"  Google Chrome ships native WebMCP off — start the session with:\n" +
+		"    sightmap browser start --chrome-flag=--enable-features=WebMCPTesting\n" +
```
chiplay, line 66: @joelgwebber Claude seems to think this is the correct flag (if needed)

joelgwebber (approved):
LGTM, with additional bits layered in:
https://github.com/sightmap/sightmap/pull/431
https://github.com/sightmap/sightkick/pull/30

Issue #159, opened 2026-08-03
SEP-0007: accept signals — composed classifications from existing entities


jurassix:
Introduces a top-level `signals:` concept: a rule that references an existing `Component`, `Request`, or `Message` by name (`ref:`) and optionally filters on that entity's own declared properties, minting a named, tagged classification. A `signals:` rule never redeclares a selector, route, or body pattern of its own — it composes what the corpus already defines, so a classification stays structurally tied to the entity it's about instead of silently drifting from it.

This issue tracks accepting [SEP-0007](https://github.com/sightmap/sightmap/blob/main/spec/seps/0007-signals.md). Implementation: #114. Depends on SEP-0005 and SEP-0006 for the `Request`/`Message` entities it references. Stays open until the SEP is accepted and the implementation lands.

joelgwebber:
One thing I'm hung up on: For things like `views:`, `requests:`, and `messages:`, the "event" is clearly implicit (navigate, network, log). But for `components:`, we've left it implicit -- is it "click", "change", etc?

We could of course specify an `event:` property, or something along those lines. But if we do, then we'll have to get into what events are possible, which feels a little downstream from sightmaps proper. Maybe worth getting into, though, if we can figure out a sensible boundary between events in the sightmap spec and tooling, and more signal/event types downstream consumers might need to add for their own purposes (eg, in subtext).

PR #445, opened 2026-09-10
feat(signals): SEP-0007 Component/View state-signal subset

joelgwebber:
Implements a small, dependency-free subset of SEP-0007 (signals) — enough to use signals as named STATE predicates (completion / availability predicates) for tooling built on top of the corpus.

Refs #159.

## What this implements

A corpus-root signals: block — {name, ref, tags?} — restricted to Component and View refs:

- ref -> a component = the boolean "that component is currently present" (evaluates exactly like a component match).
- ref -> a view = the boolean "that view's route is active" (evaluates exactly like a route match).

So a signal is the named form of a predicate the matcher already computes. Resolution is exposed as Corpus.ResolveSignal / SignalByName -> a SignalTarget (component or view). Validation reports signal-ref-unresolved and signal-ref-ambiguous (a name matching both a component and a view), plus duplicate / missing-name / missing-ref.

Kept in sync in one change: spec/v1/sightmap.schema.json, the normative spec/v1/schema.md, and the generated docs/reference/schema.md; a changeset is included (minor).

## What it deliberately does NOT implement (yet)

This is the 0005/0006-independent core of SEP-0007 — a Component ref leans only on the existing DOM component properties (SEP-0003), a View ref on nothing. Out of scope here:

- Request / Message refs — they depend on SEP-0005 (request properties) and SEP-0006 (message entity) and on stream (session) evaluation rather than current-state evaluation.
- filter: on the referenced entity's properties — a natural fast-follow (the predicate operators already exist), left out to keep this change minimal.
- The fuller temporal / range-signal machinery (windows, quantifiers, emission, absence) from the broader signal-emission proposal.

## Notes

- No runtime / matching change — signals are additive corpus data plus a resolver; a consumer turns a resolved signal into the same wait_for {query} / wait_for {route} predicate that already exists.
- go build / go vet / gofmt clean; full go test ./... green; signal_test.go covers the resolver (component/view/unknown) and every validation error path.

### DIFF
--- a/.changeset/sep-0007-state-signals.md
+++ b/.changeset/sep-0007-state-signals.md
@@ -1,0 +1,5 @@
+---
+"@sightmap/sightmap": minor
+---
+
+Add the SEP-0007 Component/View **state-signal** subset. A corpus-root `signals:` block names a component (present) or a view (route-active) as a named boolean predicate (`{name, ref, tags?}`) — the dependency-free core of SEP-0007. Resolve one with `Corpus.ResolveSignal` / `SignalByName`; validation reports `signal-ref-unresolved` and `signal-ref-ambiguous`. Request/message refs and the temporal/window machinery are intentionally out of scope.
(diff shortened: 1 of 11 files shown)


jurassix (approved):
love to see it

PR #457, opened 2026-09-17
feat(browser): browser explore drives the page toward a goal with a typed picker per step

chiplay:
## What

`sightmap browser explore --goal "..."` drives the live tab toward a goal one step at a time. Each step observes the page as the annotated component tree over one CDP connection, turns the interactive nodes into a short list of named actions, asks a picker which one to take and whether the goal is already met, acts, and waits for the page to settle. The picker only answers typed questions (pick one, yes/no); it never invents text.

The default picker is Jev via TypeSafe (`TYPESAFE_API_KEY`), which answers in about 200 ms. A step costs a few hundred milliseconds with no big-model call. `--picker anthropic` puts Claude in the same seat; `--plan` asks it once to write the goal's spec.

Stacked: the follow-up PR on `claude/browser-explore-grow` adds `--grow` (the flag is declared here and reports "not available" until that lands).

## Why in the binary

Every page command today is a fresh process that loads the corpus and opens its own CDP connection. A loop built on the CLI paid that four times a step (snapshot, click, settle polls) and ran at about 0.7 s a step. In-process, the same loop runs at 0.24 s a step on saucedemo with the model as most of the cost. Nothing new touches the daemon: the command holds one connection for its run.

## Audit: what is reused, what is new

Reused as is: `observe.Page` (extract + match + coverage in one call), `browser.Click`/`ClearAndFill`/`ScrollBy`/`EvalJSON`/`GetURL`/`NavigateAndWaitIdle`/`ClearBrowserCookies`/`ClearStorageForOrigin`/`BringToFront`, `sightmap.Load`, the corpus view matcher, `parseFlagsInterspersed`, `addLiveFlags`.

New, because the repo had none: a post-action settle (there is no DOM-quiet detector; `--wait` is a fixed sleep after `networkIdle`), an HTTP client to a model, and a goal loop. Placement follows `go/ARCHITECTURE.md`: `explore` is a live subsystem next to `observe`; `cmd_browser_explore.go` is a thin adapter.

Two findings worth their own issues: `browser click`/`fill` stall for about 5 s on some elements (the mouse hit-test and key-typed paths); this loop uses the DOM click and native-setter fill first and falls back to those. And `probe.js` re-stamps `data-sightmap-id` on every extraction, so ids must be used within the same observation (documented on `CDPDriver.Observe`).

## Numbers (same loop, both pickers; `go/explore/bench/`)

| suite | picker | goals | wall / step | model / call | Anthropic bill |
|---|---|---|---|---|---|
| saucedemo (39 components) | jev-latest | 10/10 | 0.24 s | 160 ms | $0 |
| saucedemo | claude-sonnet-5 | 10/10 | 1.16 s | 1,230 ms | $0.25 |
| books.toscrape (empty corpus, 114 raw links) | jev-latest | 10/10 | 0.30 s | 184 ms | $0 |
| books.toscrape | claude-sonnet-5 | 9/10 | 2.54 s | 1,802 ms | $0.81 |

The 12-step saucedemo checkout runs in 3.2 s. The Claude rows are a bare JSON picker, not a full agent: read them as "a big model in the same seat". Run files are in `go/explore/bench/results/`; the README there has the commands.

## Tests

- `explore`: the loop against a fake driver and picker (deterministic done, judged done, step cap, repeat guard, stale-element retry and skip, grouping with a second pick, value choice by field-name overlap and by picker, select handling), candidates and criteria, `--done-when` parsing and every check, the Jev wire format against an httptest server (auth header, question shapes, fallback on an unknown choice, retry on 5xx), the settle decision, and the suite runner with its table.
- `clitest`: the usage path fails before any CDP connection.

## Try it

```bash
export TYPESAFE_API_KEY=...
cd go/explore/bench
sightmap browser start --detach --url https://www.saucedemo.com/ --sightmap-dir saucedemo/.sightmap --profile ~/.sightmap/profiles/explore-saucedemo --port 7901 --cdp-port 7902
sightmap browser explore --sightmap-dir saucedemo/.sightmap --url https://www.saucedemo.com/ \
  --goal "Log in and put the Sauce Labs Backpack in the cart, then open the cart" \
  --done-when view=Cart --value username=standard_user --value password=secret_sauce
sightmap browser explore --bench saucedemo.json
```

Related work: [browser-use/jev-ultrafast](https://github.com/browser-use/jev-ultrafast) runs the same idea over a raw element table with operation and target heads in one request and a small text model for typed input. This loop offers sightmap names and properties as the options and takes typed values from a spec.


chiplay:
Landing in a standalone repo instead: https://github.com/sightmap/turbo (to be renamed jev-turbo), which imports github.com/sightmap/sightmap/go as a library. Nothing in sightmap changes. The two findings from this PR are filed as issues.

PR #458, opened 2026-09-17
feat(browser): browser explore --grow builds the corpus while it explores

chiplay:
Stacked on #457 (`browser explore`). Review only the top commit.

## What

`sightmap browser explore --grow` builds the corpus while it explores. On every page the loop observes:

1. Unmapped interactive nodes are grouped by their nearest stable ancestor selector (`coverage.NearestHookAncestor`, the same hooks `explain` prints) plus tag and role.
2. Jev classifies each group: buttons, links, inputs, select, nav, cards, or noise.
3. A template names it: a single control from its own text (`SearchProductsInput`), a repeated one from its container (`NavListLink`, `ProductPodLink`).
4. The scoped selector (`ul.nav-list a`) is verified against the offline matcher on that very tree, which is what `snapshot`, `coverage`, and `capture` see.
5. The component is appended to the view's YAML with `label` (and `href`) properties. Pages with no view get one named from their route pattern (`/catalogue/*/index.html` becomes `CatalogueDetail`). A selector seen on a second view is moved to `components.yaml` as a global.
6. Every write is followed by a corpus reload and `sightmap.Validate`, rolled back if that fails, and handed to the running driver so the very next step already sees the names.

Writes go through `yaml.v3` nodes rather than text surgery, so hand-written entries, comments, and `$ref`s survive.

## Audit

The repo has no YAML writer (`init` writes constants, `atlas add` untars). `grow` is the first, and it follows the stage-validate-commit shape of `atlas.Install`. Selector candidates and hooks come from `coverage`; matching from `match`; nothing in `sightmap` changed. `grow` is offline apart from the picker call and depends on `explore` only for the `Page`/`Node` types, so the dependency runs live → offline as `ARCHITECTURE.md` asks.

## Numbers

`sightmap browser explore --bench books.json --grow --repeat 3` from an empty corpus:

| repeat | goals | wall / step | coverage of visited pages | corpus after |
|---|---|---|---|---|
| 1 | 10/10 | 0.32 s | 0% → 89% | 4 views, 6 components, 4 promoted to global |
| 2 | 10/10 | 0.29 s | 100% | unchanged |
| 3 | 10/10 | 0.28 s | 100% | unchanged |

Grow work cost 1.2 s in total with seven Jev calls; `validate` and `lint --warn-only` are clean on the result. Run file: `go/explore/bench/results/books-jev-grow.json`.

Names are template names. The intended follow-up is one authoring pass over a grown corpus for better names and memory lines.

## Tests

`grow`: route patterns and view names, group naming, selector stability ranking, YAML append with comment survival and a `sightmap.Load` round trip, rollback on a failed reload, promotion to global, view creation and reuse by route, a full `OnPage` pass over a synthetic tree (two components from three orphans, nothing added on a second visit), and the noise path.


chiplay:
Landing in a standalone repo instead: https://github.com/sightmap/turbo (to be renamed jev-turbo). The grow package moved there as is.

